Issue #18, opened 2019-06-05
Add GFN-xTB energy calculator, output reader and other capabilities


andrewtarzia:
An extension of the previously added capability which added an optimization function, that calculates single point energies using the GFN software.

Also require as part of the GFN optimization/energy class:

- the ability to extract energies (total energy and/or free energy) from GFN output
- run numerical hessian calculation as part of optimization (extension of optimization class) and - - single point energy (extension of energy class)
- assign overall charge to each calculation
- assign an implicit solvent model to each calculation

andrewtarzia:
Pull request added to https://bitbucket.org/lukasturcani93/stk/src/master/ that includes all desired capability with tests

PR #25, opened 2019-06-11
Add functionality to XTB optimizer, add XTBEnergy and XTBFreeEnergy class

andrewtarzia:
Adds all functionality desired in #18.

Adds update_from_turbomole to molecule.py


stk/calculators/energy/energy_calculators.py
```diff
@@ -377,3 +382,847 @@ def energy(self, mol, conformer=-1):
...
+        Formal molecular charge.
+
+    unpaired_electrons : :class:`int`
+        Number of unpaired electrons.
+
+    use_cache : :class:`bool`
+        If ``True`` :meth:`energy` will not run twice on the same
+        molecule and conformer.
+
+    mem_ulimit : :class: `bool`
```
lukasturcani: I would prefer to rename this attribute to unlimited_memory.

stk/calculators/energy/energy_calculators.py
```diff
@@ -377,3 +382,847 @@ def energy(self, mol, conformer=-1):
...
+    output_dir : :class:`str`
+        The name of the directory into which files generated during
+        the optimization are written, if ``None`` then
+        :func:`uuid.uuid4` is used.
+
+    num_cores : :class:`int`
+        The number of cores for xTB to use. Requires appropriate setup
+        of xTB by user.
+
+    etemp : :class:`int`
```
lukasturcani: I would prefer if this was called electronic_temperature. stk uses temperature instead of temp throughout.

stk/calculators/energy/energy_calculators.py
```diff
@@ -377,3 +382,847 @@ def energy(self, mol, conformer=-1):
...
+        If ``True`` :meth:`energy` will not run twice on the same
+        molecule and conformer.
+
+    mem_ulimit : :class: `bool`
+        If ``True`` :meth:`energy` will be run without constraints on
+        the stacksize. If memory issues are encountered, this should be
+        ``True``, however this may raise issues on clusters.
+
+    total_energies : :class:`dict`
+        :class:`dict` of the total energy of each `mol`, `conformer` pair.
```
lukasturcani: This line is too long.

stk/calculators/energy/energy_calculators.py
```diff
@@ -377,3 +382,847 @@ def energy(self, mol, conformer=-1):
...
+        Key is `(mol, conformer)`
+
+    homo_lumo_gaps : :class:`dict`
+        :class:`dict` of the HOMO-LUMO gap of each `mol`, `conformer` pair.
+        Key is `(mol, conformer)`
```
lukasturcani: Needs two backticks on each side of the tuple.

lukasturcani: Needs a full stop at the end of sentence.

stk/calculators/energy/energy_calculators.py
```diff
@@ -377,3 +382,847 @@ def energy(self, mol, conformer=-1):
...
+
+    fermi_levels : :class:`dict`
+        :class:`dict` of the Fermi level of each `mol`, `conformer` pair.
+        Key is `(mol, conformer)`
```
lukasturcani: Line too long. Two backticks on each side of tuple. Needs full stop.

stk/calculators/energy/energy_calculators.py
lukasturcani: Rather than saying of each mol conformer pair (which raises the question of what mol and what conformer you are talking about), Say something like, for each 
```
:class:`.Molecule` and conformer passed to :meth:`energy`.

```

stk/calculators/energy/energy_calculators.py
```diff
@@ -377,3 +382,847 @@ def energy(self, mol, conformer=-1):
...
+    Examples
+    --------
+    .. code-block:: python
+
+        mol = StructUnit.smiles_init('NCCNCCN', ['amine'])
+        gfnxtb = XTBEnergy('/opt/gfnxtb/xtb')
+        gfnxtb.energy(mol)
+
+    Note that for :class:`.MacroMolecule` objects assembled by ``stk``
+    :class:`XTBEnergy` should usually be used after optimization with some
```
lukasturcani: Line too long.

stk/calculators/energy/energy_calculators.py
```diff
@@ -377,3 +382,847 @@ def energy(self, mol, conformer=-1):
...
+        polymer_homo_lumo_orbitals = xtbenergy.homo_lumo_orbitals[(polymer, conformer)]
+        polymer_Qonly_dipole_moments = xtbenergy.Qonly_dipole_moments[(polymer, conformer)]
+        polymer_full_dipole_moments = xtbenergy.full_dipole_moments[(polymer, conformer)]
+        polymer_Qonly_quadrupole_moments = xtbenergy.Qonly_quadrupole_moments[(polymer, conformer)]
+        polymer_QDip_quadrupole_moments = xtbenergy.QDip_quadrupole_moments[(polymer, conformer)]
+        polymer_full_quadrupole_moments = xtbenergy.full_quadrupole_moments[(polymer, conformer)]
+
+        # the total energy can be extracted at any point from the calculator
+        polymer_totalenergy = xtbenergy.total_energies[(polymer, conformer)]
+
```
lukasturcani: Lots of lines here are too long.

lukasturcani: I don't know if optimization should be discussed here. You could make the argument that before taking any energy with any calculator the structure should be optimized in some way. As a result I think it is redundant to mention this. Also, the properites and energies of any structures, not just optimized ones will be extracted. I would mentioned that this calculator can extract properties other than energy in the main section of the docstring as well, not just in the examples section.

stk/calculators/energy/energy_calculators.py
```diff
@@ -377,3 +382,847 @@ def energy(self, mol, conformer=-1):
...
+                 mem_ulimit=False):
+        """
+        Initializes a :class:`XTBEnergy` instance.
+
+        Parameters
+        ----------
+        xtb_path : :class:`str`
+            The path to the xTB executable.
+
+        gfn_version : :class:`str`
```
lukasturcani: This is an optional parameter and should be listed as such in the docstring.

stk/calculators/energy/energy_calculators.py
```diff
@@ -377,3 +382,847 @@ def energy(self, mol, conformer=-1):
...
+
+        use_cache : :class:`bool`, optional
+            If ``True`` :meth:`energy` will not run twice on the same
+            molecule and conformer.
+
+        mem_ulimit : :class: `bool`, optional
+            If ``True`` :meth:`energy` will be run without constraints on
+            the stacksize. If memory issues are encountered, this should be
+            ``True``, however this may raise issues on clusters.
+
```
lukasturcani: Check your line lengths here and the "For details:" thing.

stk/calculators/energy/energy_calculators.py
```diff
@@ -377,3 +382,847 @@ def energy(self, mol, conformer=-1):
...
+        self.qonly_dipole_moments = {}
+        self.full_dipole_moments = {}
+        self.qonly_quadrupole_moments = {}
+        self.qdip_quadrupole_moments = {}
+        self.full_quadrupole_moments = {}
+        super().__init__(use_cache=use_cache)
+
+    def _get_properties(self, mol, conformer, output_file):
+        """
+        Extracts desired properties from GFN-xTB single point energy
```
lukasturcani: The first line of a docstring should be a sentence no longer than 71 characters.

stk/calculators/energy/energy_calculators.py
```diff
@@ -377,3 +382,847 @@ def energy(self, mol, conformer=-1):
...
+        ----------
+        mol : :class:`.Molecule`
+            The :class:`.Molecule` whose energy is to be calculated.
+
+        conformer : :class:`int`, optinal
+            The conformer of `mol` to use.
+
+        output_file : :class: `str`
+            Name of output file with xTB results.
+
```
lukasturcani: Missing Returns section.

stk/calculators/energy/energy_calculators.py
```diff
@@ -377,3 +382,847 @@ def energy(self, mol, conformer=-1):
...
+        """
+        xtbext = XTBExts(output_file=output_file)
+
+        # get properties from output string
```
lukasturcani: Needs to be full sentence.

stk/calculators/energy/energy_calculators.py
```diff
@@ -377,3 +382,847 @@ def energy(self, mol, conformer=-1):
...
+        ----------
+        mol : :class:`.Molecule`
+            The :class:`.Molecule` whose energy is to be calculated.
+
+        conformer : :class:`int`, optinal
+            The conformer of `mol` to use.
+
+        Returns
+        -------
+        out_file : :class:`str`
```
lukasturcani: A variable name is not returned. dont have the "out_file : "  bit.

stk/calculators/energy/energy_calculators.py
```diff
@@ -377,3 +382,847 @@ def energy(self, mol, conformer=-1):
...
+        self.total_energies[key] = xtbext.ext_total_energy()
+        self.homo_lumo_gaps[key] = xtbext.ext_homo_lumo_gap()
+        self.fermi_levels[key] = xtbext.ext_fermi_level()
+        self.homo_lumo_orbitals[key] = xtbext.ext_homo_lumo_occ()
+        self.qonly_dipole_moments[key] = xtbext.ext_qonly_dipole_mom()
+        self.full_dipole_moments[key] = xtbext.ext_full_dipole_mom()
+        self.qonly_quadrupole_moments[key] = xtbext.ext_qonly_quadrupole_mom()
+        self.qdip_quadrupole_moments[key] = xtbext.ext_qdip_quadrupole_mom()
+        self.full_quadrupole_moments[key] = xtbext.ext_full_quadrupole_mom()
+
```
lukasturcani: Line lengths.

stk/calculators/energy/energy_calculators.py
```diff
@@ -377,3 +382,847 @@ def energy(self, mol, conformer=-1):
...
+            Name of output file with xTB results.
+
+        """
+        xyz = 'input_structure.xyz'
+        out_file = 'energy.output'
+        mol.write(xyz, conformer=conformer)
+        # Modify the memory limit.
+        if self.mem_ulimit:
+            cmd = ['ulimit -s unlimited ;']
+            # Allow multiple shell commands to be run in one subprocess.
```
lukasturcani: Too long.

stk/calculators/energy/energy_calculators.py
```diff
@@ -377,3 +382,887 @@ def energy(self, mol, conformer=-1):
...
+        self.output_dir = output_dir
+        self.num_cores = str(num_cores)
+        self.electronic_temperature = str(electronic_temperature)
+        self.solvent = solvent
+        if self.solvent is not None:
+            self.solvent = solvent.lower()
+            valid_xtb_solvent(
+                gfn_version=self.gfn_version,
+                solvent=self.solvent
+            )
```
lukasturcani: Normally the check that the input parameters are correct would be done before the state of the isntance is modified. Ie, before you do 

```python

self.xtb_path = xtb_path

```

You would have 

```python

if solvent is not None:
    solvent = solvent.lower()
    if not is_valid_solvent(gfn_version, solvent):
        raise XTBInvalidSolventError(
            f'Solvent {solvent!r} is invalid for '
            f'version {gfn_version!r}.'
        )

self.xtb_path = xtb_path
#stuff 
self.solvent = solvent

```

Note the `{solvent!r}` and `{gfn_version!r}` change to show the types in the error message.
Also note that the function now returns True or False rather than raising. The raising is done in the place where the error occurs and the function behaves more consistently. Ie returning True or False makes more sense than only returning True and raising instead of returning False, in my opinion anyway. I also think that putting `is_` in front of the function reads better and also makes it more clear what the function does, ie it makes it pretty clear that the function performs a a test which returns a bool.

stk/calculators/energy/energy_calculators.py
```diff
@@ -377,3 +382,887 @@ def energy(self, mol, conformer=-1):
...
+        self.qonly_dipole_moms = {}
+        self.full_dipole_moms = {}
+        self.qonly_quadrupole_moms = {}
+        self.qdip_quadrupole_moms = {}
+        self.full_quadrupole_moms = {}
+        super().__init__(use_cache=use_cache)
+
+    def _get_properties(self, mol, conformer, output_file):
+        """
+        Extracts properties from GFN-xTB energy calculation.
```
lukasturcani: form a* GFN-xTB

stk/calculators/energy/energy_calculators.py
```diff
@@ -377,3 +382,887 @@ def energy(self, mol, conformer=-1):
...
+
+        Parameters
+        ----------
+        mol : :class:`.Molecule`
+            The :class:`.Molecule` whose energy is to be calculated.
+
+        conformer : :class:`int`, optinal
```
lukasturcani: optinal, its a typo and also the parameter is not optional in the signature (no default is provided)

stk/calculators/energy/energy_calculators.py
```diff
@@ -377,3 +382,887 @@ def energy(self, mol, conformer=-1):
...
+            Name of output file with xTB results.
+
+        Returns
+        -------
+        None : :class:`NoneType`
+
+        """
+        xtbext = XTBExtrators(output_file=output_file)
+
+        # Get properties from output file.
```
lukasturcani: from output_file (ie you are referring to the variable directly)
or 
from the output file (ie you are referring to the concept or whatever)

stk/calculators/energy/energy_calculators.py
```diff
@@ -377,3 +382,887 @@ def energy(self, mol, conformer=-1):
...
+    def _write_and_run_command(self, mol, conformer):
+        """
+        Writes and runs the command for xTB.
+
+        Parameters
+        ----------
+        mol : :class:`.Molecule`
+            The :class:`.Molecule` whose energy is to be calculated.
+
+        conformer : :class:`int`, optinal
```
lukasturcani: typo and not even optional

stk/calculators/energy/energy_calculators.py
```diff
@@ -377,3 +382,887 @@ def energy(self, mol, conformer=-1):
...
+        # which is then closed, which minimizes the risk of
+        # unrestricting the memory limits.
+        sp.call(
+            cmd,
+            stdin=sp.PIPE,
+            stdout=f,
+            stderr=sp.PIPE,
+            shell=shell
+        )
+        f.close()
```
lukasturcani: Use `with` not close()

stk/calculators/energy/energy_calculators.py
```diff
@@ -377,3 +382,887 @@ def energy(self, mol, conformer=-1):
...
+    def energy(self, mol, conformer=-1):
+        """
+        Calculates the energy `mol` using xTB.
+
+        Parameters
+        ----------
+        mol : :class:`.Molecule`
+            The :class:`.Molecule` whose energy is to be calculated.
+
+        conformer : :class:`int`, optinal
```
lukasturcani: optinal

stk/calculators/energy/energy_calculators.py
```diff
@@ -377,3 +382,887 @@ def energy(self, mol, conformer=-1):
...
+            cmd = ['ulimit -s unlimited ;']
+            # Allow multiple shell commands to be run in one
+            # subprocess.
+            shell = True
+        else:
+            cmd = []
+            shell = False
+        cmd.append(self.xtb_path)
+        cmd.append(xyz)
+        # Set the GFN Parameterization.
```
lukasturcani: Should the P be capitalized?

stk/calculators/energy/energy_calculators.py
```diff
@@ -377,3 +382,887 @@ def energy(self, mol, conformer=-1):
...
+            cmd.append('--gbsa')
+            cmd.append(self.solvent)
+            if self.solvent_grid != 'normal':
+                cmd.append(self.solvent_grid)
+        # Write the charge section of cmd.
+        cmd.append('--chrg')
+        cmd.append(self.charge)
+        # Write the unpaired_electrons section of cmd.
+        cmd.append('--uhf')
+        cmd.append(self.unpaired_electrons)
```
lukasturcani: I would write the cmd variable quite differently and avoid all those repetitive append calls.

```python

if self.unlimited_memory:
    memory = 'ulimit -s unlimited ;'
else:
    memory = ''

if self.solvent is not None:
    if self.solvent_grid == 'normal':
        solvent_grid = ''  
    else: 
        self.solvent_grid
    solvent = f'--gbsa {self.solvent} {solvent_grid}'
else:
    solvent = ''
cmd = (
f'{memory} {self.xtb_path} {xyz} --gfn {self.gfn_version} '
f'--parallel {self.num_cores} --etemp {self.electronic_temperature} '
f'{solvent}  --chrg {self.charge} --uhf {self.unpaired_electrons}' 
)
```

stk/calculators/energy/energy_calculators.py
```diff
@@ -377,3 +382,887 @@ def energy(self, mol, conformer=-1):
...
+        # subproces is complete. This is required to be able to run the
+        # unlimited_memory_cmd and GFN calculation in one command,
+        # which is then closed, which minimizes the risk of
+        # unrestricting the memory limits.
+        sp.call(
+            cmd,
+            stdin=sp.PIPE,
+            stdout=f,
+            stderr=sp.PIPE,
+            shell=shell
```
lukasturcani: should this not just be `shell=self.unlimited_memory` ?

stk/calculators/energy/energy_calculators.py
```diff
@@ -377,3 +382,887 @@ def energy(self, mol, conformer=-1):
...
+        cmd.append('--uhf')
+        cmd.append(self.unpaired_electrons)
+
+        cmd = ' '.join(cmd)
+        f = open(out_file, 'w')
+        # Uses the shell if unlimited_memory = True and waits until the
+        # subproces is complete. This is required to be able to run the
+        # unlimited_memory_cmd and GFN calculation in one command,
+        # which is then closed, which minimizes the risk of
+        # unrestricting the memory limits.
```
lukasturcani: Not following this at all.

stk/calculators/energy/energy_calculators.py
```diff
@@ -377,3 +382,887 @@ def energy(self, mol, conformer=-1):
...
+        self.homo_lumo_gaps[key] = xtbext.homo_lumo_gap()
+        self.fermi_levels[key] = xtbext.fermi_level()
+        self.homo_lumo_orbitals[key] = xtbext.homo_lumo_occ()
+        self.qonly_dipole_moms[key] = xtbext.qonly_dipole_mom()
+        self.full_dipole_moms[key] = xtbext.full_dipole_mom()
+        self.qonly_quadrupole_moms[key] = xtbext.qonly_quadrupole_mom()
+        self.qdip_quadrupole_moms[key] = xtbext.qdip_quadrupole_mom()
+        self.full_quadrupole_moms[key] = xtbext.full_quadrupole_mom()
+
+    def _write_and_run_command(self, mol, conformer):
```
lukasturcani: I think if your function has the word "and" in its name, it's a good indicator that two functions should be used in succession instead.

stk/calculators/energy/energy_calculators.py
andrewtarzia: # Uses the shell if unlimited_memory = True to run multiple commands.

stk/calculators/energy/energy_calculators.py
```diff
@@ -377,3 +382,887 @@ def energy(self, mol, conformer=-1):
...
+        solvent.
+        Can be one of ``'normal'``, ``'tight'``, ``'verytight'``
+        or ``'extreme'``
+        For details see
+        https://xtb-docs.readthedocs.io/en/latest/gbsa.html
+
+    charge : :class:`int`
+        Formal molecular charge.
+
+    unpaired_electrons : :class:`int`
```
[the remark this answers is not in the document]
andrewtarzia: replaced in all .py

stk/calculators/energy/energy_calculators.py
```diff
@@ -377,3 +382,887 @@ def energy(self, mol, conformer=-1):
...
+        Number of unpaired electrons.
+
+    use_cache : :class:`bool`
+        If ``True`` :meth:`energy` will not run twice on the same
+        molecule and conformer.
+
+    unlimited_memory : :class: `bool`
+        If ``True`` :meth:`energy` will be run without constraints on
+        the stacksize. If memory issues are encountered, this should be
+        ``True``, however this may raise issues on clusters.
```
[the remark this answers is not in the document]
andrewtarzia: indeed

stk/calculators/energy/energy_calculators.py
```diff
@@ -377,3 +382,887 @@ def energy(self, mol, conformer=-1):
...
+    qdip_quadrupole_moms : :class:`dict`
+        :class:`dict` of the `q+dip` quadrupole moment of
+        :class:`.Molecule` and conformer passed to :meth:`energy`.
+        Key is ``(mol, conformer)``.
+
+    full_quadrupole_moms : :class:`dict`
+        :class:`dict` of the `full` quadrupole moment of
+        :class:`.Molecule` and conformer passed to :meth:`energy`.
+        Key is ``(mol, conformer)``.
+
```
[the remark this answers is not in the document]
andrewtarzia: changed mom to moment throughout

stk/calculators/energy/energy_calculators.py
```diff
@@ -377,3 +382,887 @@ def energy(self, mol, conformer=-1):
...
+    Examples
+    --------
```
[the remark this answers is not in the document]
andrewtarzia: I have used the UFF optimizer here to avoid the double up on documentation.

stk/calculators/energy/energy_calculators.py
```diff
@@ -377,3 +382,887 @@ def energy(self, mol, conformer=-1):
...
+            (polymer, conformer)
+        ]
+        p_full_quadrupole_moms = xtb.full_quadrupole_moms[
+            (polymer, conformer)
+        ]
+
+        # the total energy can be extracted at any point from the
+        # calculator
+        polymer_totalenergy = xtb.total_energies[(polymer, conformer)]
+
```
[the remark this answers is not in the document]
andrewtarzia: Done. and I agree completely

stk/calculators/energy/energy_calculators.py
```diff
@@ -377,3 +382,887 @@ def energy(self, mol, conformer=-1):
...
+                 unlimited_memory=False):
+        """
+        Initializes a :class:`XTBEnergy` instance.
+
+        Parameters
+        ----------
+        xtb_path : :class:`str`
+            The path to the xTB executable.
+
+        gfn_version : :class:`str`, optional
```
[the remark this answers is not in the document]
andrewtarzia: changed to int in all functions and docs

stk/calculators/energy/energy_calculators.py
andrewtarzia: Very good points @lukasturcani

stk/calculators/energy/energy_calculators.py
andrewtarzia: from* ;)

stk/calculators/energy/energy_calculators.py
andrewtarzia: this is a good point. I try to

andrewtarzia: Normally I would agree, however, I think these two processes are so inherently linked and to require two functions just adds function calls and doesn't improve the readability of the code. If you disagree, then I can split them, but for now I am leaving them as is.

stk/calculators/energy/energy_calculators.py
andrewtarzia: brilliant

stk/calculators/energy/energy_calculators.py
lukasturcani: I dont think the processes are inherently linked. I think they are only linked by the conext of  their calling funciton. IE writing does not depending on running the command. The execution of a subprocess call is also independent of the write being done first. Now of course, in the context of an energy calculation the write needs to happen first, followed by a subprocess call, which is why the energ() function is the thing that links these two functions within a context.

```python

mol.write(...)
self._run_cmd()

``` 

reads better.

lukasturcani: Also write_and_run_cmd() raises the question "write what?" whereas this makes it clear that the mol is getting written. I would say its cool to put mol.write() within _run_cmd, but just call it _run_cmd() dont call it write_and_run_cmd()

andrewtarzia: I agree with this and realized the confusion due to how I wrote the code.

stk/calculators/energy/energy_calculators.py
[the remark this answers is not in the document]
lukasturcani: Use an optimizer sequence then. I think an example in the documentation should be usable.

stk/calculators/optimization/optimizers.py
```diff
@@ -662,88 +672,457 @@ class GFNXTB(Optimizer):
...
+
+        """
+        neg_freq = False
+        xtbext = XTBExtractor(output_file=output_file)
+        value = xtbext.frequencies()
+        # Check for one negative frequency, excluding the first
+        # 6 frequencies.
+        if min(value[7:]) < 0:
+            neg_freq = True
+        return neg_freq
```
lukasturcani: This function can be written more cleanly

```python

        xtbext = XTBExtractor(output_file=output_file)
        value = xtbext.frequencies()
        # Check for one negative frequency, excluding the first
        # 6 frequencies.
        return min(value[6:]) < 0
```

stk/calculators/optimization/optimizers.py
```diff
@@ -662,88 +672,457 @@ class GFNXTB(Optimizer):
...
+
+        Parameters
+        ----------
+        output_file : :class:`str`
+            Name of output file with xTB results.
+
+        Returns
+        -------
+        :class:`bool`
+            Returns `True` if a negative frequency is present. Raises
```
lukasturcani: True needs two backticks on each side

stk/calculators/optimization/optimizers.py
```diff
@@ -662,88 +672,457 @@ class GFNXTB(Optimizer):
...
+            errors if optimization did not converge.
+
+        """
+        if output_file is None:
+            # No simulation has been run.
+            return True
+        # If convergence is achieved, then .xtboptok should exist.
+        if os.path.exists('.xtboptok'):
+            # Check for negative frequencies in output file in max_run
+            # is not None..
```
lukasturcani: Two full stops here.

Also im guessing this will be rewritten when max_runs is changed.

stk/calculators/optimization/optimizers.py
```diff
@@ -662,88 +672,457 @@ class GFNXTB(Optimizer):
...
+        Parameters
+        ----------
+        output_file : :class:`str`
+            Name of output file with xTB results.
+
+        Returns
+        -------
+        :class:`bool`
+            Returns `True` if a negative frequency is present. Raises
+            errors if optimization did not converge.
```
lukasturcani: The returns section should only documented the return values. Comments about what the function raises should be left to the Raises section below.

stk/calculators/optimization/optimizers.py
```diff
@@ -662,88 +672,457 @@ class GFNXTB(Optimizer):
...
+            # Return True if there exists at least one.
+            if self.max_runs is not None:
+                return self._check_neg_frequencies(output_file)
+            else:
+                return False
+        elif os.path.exists('NOT_CONVERGED'):
+            XTBConvergenceError('Optimization not converged.')
+        else:
+            XTBOptimizerError('Optimization failed to complete')
+
```
lukasturcani: Neither of these errors is raised.

stk/calculators/optimization/optimizers.py
```diff
@@ -662,88 +672,457 @@ class GFNXTB(Optimizer):
...
+    def _run_xtb(self, xyz, out_file):
+        """
+        Runs the command for GFN-xTB using subprocess.
```
lukasturcani: The use of subprocess is an implementation detail and should not be mentioned in the docstring.

stk/calculators/energy/energy_calculators.py
```diff
@@ -377,3 +384,475 @@ def energy(self, mol, conformer=-1):
...
+            xtbext.qdip_quadrupole_moment()
+        self.full_quadrupole_moments[key] = \
+            xtbext.full_quadrupole_moment()
+        if self.calculate_free_energy:
+            self.total_free_energies[key] = xtbext.total_free_energy()
+            self.frequencies[key] = xtbext.frequencies()
+
+    def _run_xtb(self, xyz, out_file):
+        """
+        Runs the command for GFN-xTB.
```
lukasturcani: Runs GFN-xTB.

would be better.

stk/calculators/optimization/optimizers.py
```diff
@@ -662,88 +672,457 @@ class GFNXTB(Optimizer):
...
+
+        Parameters
+        ----------
+        xyz : :class:`str`
+            The name of the input structure `.xyz` file.
```
lukasturcani: Two backticks on each side of .xyz

stk/calculators/optimization/optimizers.py
```diff
@@ -662,88 +672,457 @@ class GFNXTB(Optimizer):
...
+            optimization = f'--opt {self.opt_level}'
+        else:
+            # Do optimization and check hessian.
+            optimization = f'--ohess {self.opt_level}'
+
+        if self.solvent is not None:
+            if self.solvent_grid == 'normal':
+                solvent_grid = ''
+            else:
+                solvent_grid = ''
```
lukasturcani: If and else both do the same thing.

stk/calculators/optimization/optimizers.py
```diff
@@ -662,88 +672,457 @@ class GFNXTB(Optimizer):
...
+                    self.incomplete.add((mol, conformer))
+                    logging.warning(
+                        'Small negative frequencies present in '
+                        f'{mol.name} conformer {conformer}.'
+                    )
+                    return False
+                else:
+                    # Calculation is complete.
+                    return True
+        return None
```
lukasturcani: Don't write return None, delete this line entirely, it is completely redundant

Not writing it  does the exact same thing

even 

```
return 
```

followeding by nothing is the same as return None.

stk/molecular/molecules/molecule.py
```diff
@@ -1114,6 +1114,77 @@ def update_from_xyz(self, path, conformer=-1):
         new_coords = np.array(new_coords).T
         self.set_position_from_matrix(new_coords, conformer=conformer)
 
+    def update_from_turbomole(self, path, conformer=-1):
+        """
+        Updates molecular structure to match a Turbomole ``.coord`` file.
```
lukasturcani: This line is too long.

stk/molecular/molecules/molecule.py
```diff
@@ -1114,6 +1114,77 @@ def update_from_xyz(self, path, conformer=-1):
...
+
+        Note that coordinates in ``.coord`` files are given in Bohr.
+
+        Parameters
+        ----------
+        path : :class:`str`
+            The full path of the ``.mol`` file from which the structure
```
lukasturcani: I thought it was a .coord file?

stk/molecular/molecules/molecule.py
```diff
@@ -1114,6 +1114,77 @@ def update_from_xyz(self, path, conformer=-1):
...
+
+        if conformer == -1:
+            conformer = self.mol.GetConformer(conformer).GetId()
+
+        with open(path, 'r') as f:
+            _, *content, __ = f.readlines()
+
+        # Check the atom count is correct.
+        num_atoms = self.mol.GetNumAtoms()
+        if int(len(content)) != num_atoms:
```
lukasturcani: Why are you putting the return vlaue of len into int?

stk/molecular/molecules/molecule.py
```diff
@@ -1114,6 +1114,77 @@ def update_from_xyz(self, path, conformer=-1):
...
+            raise RuntimeError(
+                f'The number of atoms in the xyz file, {len(content)}, '
```
lukasturcani: This line is too long.

stk/molecular/molecules/molecule.py
```diff
@@ -1114,6 +1114,77 @@ def update_from_xyz(self, path, conformer=-1):
...
+                    f'Atom {i} element does not match file.'
+                )
+
+            new_coords.append([float(i)*bohr_to_ang for i in coords])
+
+        # Check that the correct number of atom
+        # lines was present in the file.
+        if i+1 != num_atoms:
+            raise RuntimeError(
+                f'The number of atom lines in the xyz file, {i+1}, '
```
lukasturcani: I thought it was a .coord file?

stk/molecular/molecules/molecule.py
lukasturcani: Also, i thought it was a .coord file?

stk/utilities/utilities.py
```diff
@@ -1143,10 +1146,438 @@ def vector_theta(vector1, vector2):
...
         return 0.0
     return np.arccos(numerator/denominator)
+
+
+class XTBInvalidSolventError(Exception):
+    ...
+
+
+def is_valid_xtb_solvent(gfn_version, solvent):
+    '''Check if solvent is valid for the given GFN version.
```
lukasturcani: The first line of a docstring should be on the line below the quotes.

Also in stk three """ are used not '''

stk/utilities/utilities.py
```diff
@@ -1143,10 +1146,438 @@ def vector_theta(vector1, vector2):
...
+
+    Parameters
+    ----------
+    gfn_version : :class:`str`
+        GFN parameterization version. Can be: 0, 1 or 2.
```
lukasturcani: The number strings need correct markup.

stk/utilities/utilities.py
```diff
@@ -1143,10 +1146,438 @@ def vector_theta(vector1, vector2):
...
+    -------
+    :class:`bool`
+        ``True`` if solvent is valid.
+
+    See Also
+    --------
+    # https://xtb-docs.readthedocs.io/en/latest/gbsa.html
+
+    '''
+    if str(gfn_version) == '0':
```
lukasturcani: The docstring says that this parameter is already a string, but here you are converting it into a string. Which one is wrong?

stk/utilities/utilities.py
```diff
@@ -1143,10 +1146,438 @@ def vector_theta(vector1, vector2):
...
+        return False
+    elif str(gfn_version) == '1':
+        valid_solvents = [
```
lukasturcani: valid solvents should be a set not a list, because you use it to check for membership.

stk/utilities/utilities.py
```diff
@@ -1143,10 +1146,438 @@ def vector_theta(vector1, vector2):
...
+            'acetone', 'acetonitrile', 'benzene',
+            'CH2Cl2'.lower(), 'CHCl3'.lower(), 'CS2'.lower(),
+            'DMSO'.lower(), 'ether', 'H2O'.lower(),
+            'methanol', 'THF'.lower(), 'toluene'
+        ]
+        return solvent in valid_solvents
+    elif str(gfn_version) == '2':
+        valid_solvents = [
```
lukasturcani: valid solvents hould be a set not a list.

stk/utilities/utilities.py
```diff
@@ -1143,10 +1146,438 @@ def vector_theta(vector1, vector2):
...
+            'n-hexane'.lower(), 'THF'.lower(), 'toluene'
+        ]
+        return solvent in valid_solvents
+
+
+class XTBExtractor:
+    """
+    Methods for extracting properties from GFN-xTB output.
+
+    Parameters
```
lukasturcani: This is a class docstring, did you mean Attributes?

stk/utilities/utilities.py
```diff
@@ -1143,10 +1146,438 @@ def vector_theta(vector1, vector2):
...
+            'CHCl3'.lower(), 'CS2'.lower(), 'DMF'.lower(),
+            'DMSO'.lower(), 'ether', 'H2O'.lower(), 'methanol',
+            'n-hexane'.lower(), 'THF'.lower(), 'toluene'
+        ]
+        return solvent in valid_solvents
+
+
+class XTBExtractor:
+    """
+    Methods for extracting properties from GFN-xTB output.
```
lukasturcani: How about 

Extracts properties from GFN-xTB output files.

?

stk/utilities/utilities.py
```diff
@@ -1143,10 +1146,438 @@ def vector_theta(vector1, vector2):
...
+
+    Parameters
+    ----------
+    output_file : :class:`str`
+        Output file to extract properties from.
+
+    """
+    def __init__(self, output_file):
```
lukasturcani: Needs a docstring.

stk/utilities/utilities.py
```diff
@@ -1143,10 +1146,438 @@ def vector_theta(vector1, vector2):
...
+        output_string = open(output_file, 'r').readlines()
```
lukasturcani: Always, always, always use "with"
Here you did not close the file...

stk/utilities/utilities.py
```diff
@@ -1143,10 +1146,438 @@ def vector_theta(vector1, vector2):
...
+
+        Formatting based on latest version of xTB (190418)
+        Example line:
+        "          | TOTAL ENERGY              -76.260405590154 Eh   |"
+
+        Returns
+        -------
+        :class:`float`
+            Total energy in a.u.
+        """
```
lukasturcani: empty line before the """ please.

stk/utilities/utilities.py
```diff
@@ -1143,10 +1146,438 @@ def vector_theta(vector1, vector2):
...
+        value = None
+
+        # Regular expression for numbers.
+        nums = re.compile(r"[+-]?\d+(?:\.\d+)?(?:[eE][+-]?\d+)?")
+        for line in reversed(self.output_string):
+            if '          | TOTAL ENERGY  ' in line:
+                value = nums.search(line.rstrip()).group(0)
+                break
+
+        return float(value)
```
lukasturcani: This function can be written more cleanly

```python

        # Regular expression for numbers.
        nums = re.compile(r"[+-]?\d+(?:\.\d+)?(?:[eE][+-]?\d+)?")
        for line in reversed(self.output_string):
            if '          | TOTAL ENERGY  ' in line:
                return float(nums.search(line.rstrip()).group(0))
```

stk/utilities/utilities.py
```diff
@@ -1143,10 +1146,438 @@ def vector_theta(vector1, vector2):
...
+    def __init__(self, output_file):
+        output_string = open(output_file, 'r').readlines()
+        self.output_string = output_string
+
+    def total_energy(self):
+        """
+        Extracts total energy in a.u. from xTB output.
+
+        Formatting based on latest version of xTB (190418)
+        Example line:
```
lukasturcani: Remove the example line bit imo, in all other methods too

stk/utilities/utilities.py
lukasturcani: This applies to all the other similar methods in this class

stk/utilities/utilities.py
```diff
@@ -1143,10 +1146,438 @@ def vector_theta(vector1, vector2):
...
+            if '             Fermi-level        ' in line:
+                part2 = line.split('Eh')
+                value = nums.search(part2[1].rstrip()).group(0)
+                break
+
+        return float(value)
+
+    def qonly_dipole_moment(self):
+        """
+        Extracts `q only` dipole moment in Debye from xTB output.
```
lukasturcani: q only has wrong markup

stk/utilities/utilities.py
```diff
@@ -1143,10 +1146,438 @@ def vector_theta(vector1, vector2):
...
+        value = None
+
+        sample_set = []
+        for i, line in enumerate(self.output_string):
+            if 'molecular dipole:' in line:
+                sample_set = self.output_string[i+2].rstrip()
+
+        # get values from line
+        if 'q only:' in sample_set:
+            x, y, z = [i for i in sample_set.split(':')[1].split(' ')
```
lukasturcani: this list comprehension is incorrectly indented and can be written more cleanly by creating the float right away  

```python

return [
    float(i)
    for i in sample.set(':')[1].split(' ') if i
]

```
plus you can return right away as well

lukasturcani: Note that you dont need to compare to empty string because any 
if somestring

is true, for any string that is not the empty string and false for the empty string.

stk/utilities/utilities.py
```diff
@@ -1143,10 +1146,438 @@ def vector_theta(vector1, vector2):
...
+                       if i != '']
+
+        value = [float(x), float(y), float(z)]
+
+        return value
+
+    def full_dipole_moment(self):
+        """
+        Extracts `full` dipole moment in Debye from xTB output.
```
lukasturcani: wrong markup around full

stk/utilities/utilities.py
```diff
@@ -1143,10 +1146,438 @@ def vector_theta(vector1, vector2):
...
+
+        Formatting based on latest version of xTB (190418)
+        Example line:
+        " q only:       -0.033      -0.081      -0.815"
+
+        Returns
+        -------
+        :class:`list` of :class:`float`
+            Components of dipole moment as ``[`x`, `y`, `z`]`` in a
+            list of length 3.
```
lukasturcani: wrong markup in the list

stk/utilities/utilities.py
```diff
@@ -1143,10 +1146,438 @@ def vector_theta(vector1, vector2):
...
+
+        Formatting based on latest version of xTB (190418)
+        Example line:
+        "   full:       -0.684       0.122      -1.071       3.245"
+
+        Returns
+        -------
+        :class:`list` of :class:`float`
+            Components of dipole moment and total magnitude as
+            ``[`x`, `y`, `z`, `total`]`` in a list of length 4.
```
lukasturcani: missing markup around list and wrong markup within the list example

stk/utilities/utilities.py
```diff
@@ -1143,10 +1146,438 @@ def vector_theta(vector1, vector2):
...
+        for i, line in enumerate(self.output_string):
+            if 'molecular dipole:' in line:
+                sample_set = self.output_string[i+3].rstrip()
+
+        # get values from line
+        if 'full:' in sample_set:
+            samp = sample_set.split(':')[1].split(' ')
+            x, y, z, m = [i for i in samp if i != '']
+
+        value = [float(x), float(y), float(z), float(m)]
```
lukasturcani: again, use the list comprehension more efficiently here and get rid of comparison to emptyu string.

stk/calculators/energy/energy_calculators.py
```diff
@@ -377,3 +384,475 @@ def energy(self, mol, conformer=-1):
...
+        conformer passed to :meth:`energy`. The key has the
+        form ``(mol, conformer)``.
+
+    homo_lumo_orbitals : :class:`dict`
+        :class:`dict` of the HOMO-LUMO orbital properties of
+        each :class:`.Molecule` and conformer passed to :meth:`energy`.
+        The key has the form ``(mol, conformer)``.
+
+    qonly_dipole_moments : :class:`dict`
+        :class:`dict` of the `q only` dipole momentent of
```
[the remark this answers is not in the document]
andrewtarzia: fixed here and in utilities.

stk/calculators/energy/energy_calculators.py
```diff
@@ -377,3 +384,475 @@ def energy(self, mol, conformer=-1):
...
+            out_file = 'energy.output'
+            mol.write(xyz, conformer=conformer)
+            self._run_xtb(xyz=xyz, out_file=out_file)
+            self._get_properties(
+                mol=mol,
+                conformer=conformer,
+                output_file=out_file
+            )
+        finally:
+            os.chdir(init_dir)
```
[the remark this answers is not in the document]
andrewtarzia: excellent point. thank you

stk/utilities/utilities.py
andrewtarzia: could not agree more.

stk/calculators/optimization/optimizers.py
```diff
@@ -773,19 +1178,18 @@ def optimize(self, mol, conformer=-1):
...
             os.chdir(init_dir)
+
+        if result is None:
+            self.incomplete.add((mol, conformer))
+            logging.warning(
+                'Negative frequencies present in '
+                f'{mol.name} conformer {conformer} '
+                f'after {self.max_runs} optimizations '
+                'completed.'
+            )
```
lukasturcani: In this function, because (mol, conformer) is used so frequently, i would reccomend placing it into a variable.

stk/calculators/optimization/optimizers.py
```diff
@@ -648,102 +650,501 @@ def optimize(self, mol, conformer=-1):
...
+        ----------
+        mol : :class:`.Molecule`
+            The molecule to be optimized.
+
+        conformer : :class:`int`
+            The conformer to use.
+
+        Returns
+        -------
+        None : :class:`NoneType`
```
lukasturcani: I belive this functions returns True or False

stk/calculators/optimization/optimizers.py
```diff
@@ -648,102 +650,501 @@ def optimize(self, mol, conformer=-1):
...
+                    # case, exit optimization loop and warn.
+                    self.incomplete.add((mol, conformer))
+                    logging.warning(
+                        'Small negative frequencies present in '
+                        f'{mol.name} conformer {conformer}.'
+                    )
+                    return False
+                else:
+                    # Calculation is complete.
+                    return True
```
lukasturcani: I think the nesting here can be made a little bit clearer

```python

            if opt_incomplete:
                if coord_exists:
                    # The calculation is incomplete.
                    # Update mol from xtbhess.coord and continue.
                    mol.update_from_turbomole(
                        path=coord_file,
                        conformer=conformer
                    )
                else:
                    # Update mol from xtbopt.xyz.
                    mol.update_from_xyz(
                        path=output_xyz,
                        conformer=conformer
                    )
                    # If the negative frequencies are small, then GFN
                    # may not produce the restart file. If that is the
                    # case, exit optimization loop and warn.
                    self.incomplete.add((mol, conformer))
                    logging.warning(
                        'Small negative frequencies present in '
                        f'{mol.name} conformer {conformer}.'
                    )
                    return False
            else:
                break
        return True

```

Note the addition of return True at the end of the function. and the insertion of break in the loop.

stk/calculators/optimization/optimizers.py
```diff
@@ -773,19 +1178,18 @@ def optimize(self, mol, conformer=-1):
...
-                self.gfnxtb_path,
-                xyz,
-                '-opt',
-                '--parallel',
-                self.num_cores
-            ]
-            sp.run(cmd, stdin=sp.PIPE, stdout=sp.PIPE, stderr=sp.PIPE)
-            output_xyz = join(output_dir, 'xtbopt.xyz')
-            mol.update_from_xyz(path=output_xyz, conformer=conformer)
+            result = self._run_optimizations(mol, conformer)
```
lukasturcani: how about changing result to "complete" 
which will always be a True or False now given the _run_optimizeations changes.

stk/calculators/optimization/optimizers.py
```diff
@@ -773,19 +1178,18 @@ def optimize(self, mol, conformer=-1):
...
         finally:
             os.chdir(init_dir)
+
+        if result is None:
```
lukasturcani: This could be "if not complete:"
which would read pretty well imo

stk/calculators/optimization/optimizers.py
lukasturcani: The error message will have to be changed however, because the function can exit early, ie when runs performed is less than max_runs and still return False.

lukasturcani: Maybe changing it to something along the lines of "negative frequencies present in final structure of {mol.name} conformer {conformer}"

lukasturcani: If you really want to make the user know how many runs were done, you can always done always use the logger (maybe info level?) if you think its necessary / beneficial

stk/utilities/utilities.py
```diff
@@ -1143,10 +1146,371 @@ def vector_theta(vector1, vector2):
...
+    """
+    if gfn_version == 0:
+        return False
+    elif gfn_version == 1:
+        valid_solvents = set([
+            'acetone', 'acetonitrile', 'benzene',
+            'CH2Cl2'.lower(), 'CHCl3'.lower(), 'CS2'.lower(),
+            'DMSO'.lower(), 'ether', 'H2O'.lower(),
+            'methanol', 'THF'.lower(), 'toluene'
+        ])
```
lukasturcani: Dont make a set from a list, because you are still making the list first, which comes at both a computational and memory cost, You can create a set directly with
```python
        valid_solvents = {
            'acetone', 'acetonitrile', 'benzene',
            'CH2Cl2'.lower(), 'CHCl3'.lower(), 'CS2'.lower(),
            'DMSO'.lower(), 'ether', 'H2O'.lower(),
            'methanol', 'THF'.lower(), 'toluene'
        }

```

lukasturcani: Those are curly brackets (its not obvious on my github). bascially curly brackets where values are separated by commas and ":" is not present means set. Ofc if you add ":" you make a dict. {} defaults to empty dict, so if you want to make an empty set use "set()"

stk/utilities/utilities.py
```diff
@@ -1143,10 +1146,371 @@ def vector_theta(vector1, vector2):
...
+
+class XTBExtractor:
+    """
+    Extracts properties from GFN-xTB output files.
+
+    Attributes
+    ----------
+    output_file : :class:`str`
+        Output file to extract properties from.
+
```
lukasturcani: output string is not documented.

stk/utilities/utilities.py
```diff
@@ -1143,10 +1146,371 @@ def vector_theta(vector1, vector2):
...
+        Parameters
+        ----------
+        output_file : :class:`str`
+            Output file to extract properties from.
+
+        """
+        self.output_file = output_file
+        with open(self.output_file, 'r') as f:
+            self.output_string = f.readlines()
```
lukasturcani: should it be called output_string if it's actually a list?

stk/utilities/utilities.py
```diff
@@ -1143,10 +1146,371 @@ def vector_theta(vector1, vector2):
...
+    def qonly_dipole_moment(self):
+        """
+        Extracts q only dipole moment in Debye from xTB output.
+
+        Formatting based on the 190418 version of xTB.
+
+        Returns
+        -------
+        :class:`list` of :class:`float`
+            Components of dipole moment as [x, y, z] in a
```
lukasturcani: two backticks on each side of list

lukasturcani: probably dont need to state that its length 3 if you are giving the list directly.

stk/utilities/utilities.py
```diff
@@ -1143,10 +1146,371 @@ def vector_theta(vector1, vector2):
...
+        """
+        Extracts full dipole moment in Debye from xTB output.
+
+        Formatting based on the 190418 version of xTB.
+
+        Returns
+        -------
+        :class:`list` of :class:`float`
+            Components of dipole moment and total magnitude as
+            [x, y, z, total] in a :class:`list` of length 4.
```
lukasturcani: 2 backticks on each end of list and i think statement on length is redundant.

stk/utilities/utilities.py
```diff
@@ -1143,10 +1146,371 @@ def vector_theta(vector1, vector2):
...
+        """
+        Extracts q only traceless quadrupole moment in Debye.
+
+        Formatting based on the 190418 version of xTB.
+
+        Returns
+        -------
+        :class:`list` of :class:`float`
+            Components of quadrupole moment as
+            [xx, xy, xy, xz, yz, zz] in a :class:`list` of length 6.
```
lukasturcani: as above

stk/utilities/utilities.py
```diff
@@ -1143,10 +1146,371 @@ def vector_theta(vector1, vector2):
...
+        """
+        Extracts q+dip traceless quadrupole moment in Debye.
+
+        Formatting based on the 190418 version of xTB.
+
+        Returns
+        -------
+        :class:`list` of :class:`float`
+            Components of quadrupole moment as
+            [xx, xy, xy, xz, yz, zz] in a :class:`list` of length 6.
```
lukasturcani: as above

stk/utilities/utilities.py
```diff
@@ -1143,10 +1146,371 @@ def vector_theta(vector1, vector2):
...
+
+        """
+
+        sample_set = []
+        for i, line in enumerate(self.output_string):
+            if 'molecular quadrupole (traceless):' in line:
+                sample_set = self.output_string[i+3].rstrip()
+
+        # get values from line
```
lukasturcani: Full sentence please

stk/utilities/utilities.py
```diff
@@ -1143,10 +1146,371 @@ def vector_theta(vector1, vector2):
...
+        :class:`list` of :class:`float`
+            Components of quadrupole moment as
+            [xx, xy, xy, xz, yz, zz] in a :class:`list` of length 6.
+
+        """
+
+        sample_set = []
+        for i, line in enumerate(self.output_string):
+            if 'molecular quadrupole (traceless):' in line:
+                sample_set = self.output_string[i+3].rstrip()
```
lukasturcani: do we need to keep iterating through these lines after the value has been found? I would assume so since maybe you want the last line, but maybe worth a comment? (applies to other functions too)

stk/utilities/utilities.py
```diff
@@ -1143,10 +1146,371 @@ def vector_theta(vector1, vector2):
...
+        :class:`dict`
+            :class:`dict` with items of the form
+            'HOMO':(#, occupation, energy) and
+            'LUMO':(#, occupation, energy).
+
+        """
+
+        for line in reversed(self.output_string):
+            if '(HOMO)' in line:
+                split_line = [i for i in line.rstrip().split(' ')
```
lukasturcani: indentation of list is off here

lukasturcani: and comparison to null string is redundant

stk/utilities/utilities.py
```diff
@@ -1143,10 +1146,371 @@ def vector_theta(vector1, vector2):
...
+            if '(LUMO)' in line:
+                split_line = [i for i in line.rstrip().split(' ')
+                              if i != '']
+                # The line is:
+                #   Number, energy (Ha), energy (ev), label
+                # Extract:
+                #   Number, energy (eV)
+                lumo_val = [
+                    int(split_line[0]),
+                    float(0),
```
lukasturcani: this is a bit weird. why not just 0 or "0." if you really want that float?

stk/utilities/utilities.py
```diff
@@ -1143,10 +1146,371 @@ def vector_theta(vector1, vector2):
...
+                switch = False
+            if 'eigval :' in line and switch is True:
+                samp = line.rstrip().split(':')[1].split(' ')
+                split_line = [i for i in samp if i != '']
+                for freq in split_line:
+                    frequencies.append(freq)
+
+        value = [float(i) for i in frequencies]
+
+        return value
```
lukasturcani: remove redundant variable assignment

stk/calculators/energy/energy_calculators.py
```diff
@@ -377,3 +385,469 @@ def energy(self, mol, conformer=-1):
...
+        conformer passed to :meth:`energy`. The key has the
+        form ``(mol, conformer)``.
+
+    homo_lumo_orbitals : :class:`dict`
+        :class:`dict` of the HOMO-LUMO orbital properties of
+        each :class:`.Molecule` and conformer passed to :meth:`energy`.
+        The key has the form ``(mol, conformer)``.
+
+    qonly_dipole_moments : :class:`dict`
+        :class:`dict` of the q only dipole momentent of
```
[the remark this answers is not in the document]
andrewtarzia: find and replace of mom -> moment....

stk/calculators/optimization/optimizers.py
```diff
@@ -648,102 +650,501 @@ def optimize(self, mol, conformer=-1):
...
+        # The conformer must be set to check the `incomplete`
+        # attribute.
+        conformer = polymer.mol.GetConformer(-1).GetId()
+
+        xtb_crude.optimize(mol=polymer, conformer=conformer)
+        if (polymer, conformer) in xtb_crude.incomplete:
+            xtb_normal.optimize(mol=polymer, conformer=conformer)
+            if (polymer, conformer) in xtb_normal.incomplete:
+                xtb_vtight.optimize(mol=polymer, conformer=conformer)
+
```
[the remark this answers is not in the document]
andrewtarzia: I like this v much

stk/calculators/optimization/optimizers.py
andrewtarzia: This was old, right?

stk/calculators/optimization/optimizers.py
andrewtarzia: please check the changes here

stk/calculators/optimization/optimizers.py
andrewtarzia: this loop does not update the mol from the xyz if opt_incomplete is False. That was the reason for the weird nesting.

stk/calculators/optimization/optimizers.py
andrewtarzia: please check the logic I have used in this funciton and the one above

stk/utilities/utilities.py
andrewtarzia: nah, there should be a break there.

stk/calculators/optimization/optimizers.py
lukasturcani: The way its written at the moment, it does update from xyz when opt_incomplete is False and coord_exists is True

stk/utilities/utilities.py
```diff
@@ -1143,10 +1146,450 @@ def vector_theta(vector1, vector2):
...
+    Extracts properties from GFN-xTB output files.
+
+    Attributes
+    ----------
+    output_file : :class:`str`
+        Output file to extract properties from.
+
+    output_list : :class:`list`
+        :class:`list` of all lines in as :class:`str` in output file.
+
```
lukasturcani: This class is missing the documentation for all the extracted attributes.
once those are documented here, you can remove the things like 

```
        Gives components of quadrupole moment as
        ``[x, y, z]`` in a :class:`list`.

```

in the extract method's documentation. Instead you can just write

```

Updates :attr:`qonly_quadrupole_moment`.

```

Which will prevent the need to document the same thing twice in in different places (which leads to a high chance that one  of them will forget to be updated when the other changes)

stk/calculators/optimization/optimizers.py
```diff
@@ -648,102 +650,503 @@ def optimize(self, mol, conformer=-1):
...
+    :meth:`optimize` will check that the structure is appropriately
+    optimized (i.e. convergence is obtained and no negative vibrational
+    frequencies are present) and continue optimizing a structure (up to
+    :attr:`max_runs` times) until this is achieved. This loop, by
+    default, will be performed at the same :attr:`opt_level`. The
+    following examples shows how a user may optimize structures with
+    tigher convergence criteria (i.e. different :attr:`opt_level`)
+    until the structure is sufficiently optimized. Furthermore, the
+    calculation of the Hessian can be turned off using
+    :attr:`max_runs` = 1 and :attr:`calculate_hessian`=False.
```
lukasturcani: This needs to be written as a full sentence.

```
Furthermore, the calculation of the Hessian can be turned off by setting :attr:`max_runs` to ``1`` and 
:attr:`calculate_hessian` to ``False``.

```

This is because the syntax highlight will be off in the way you have written it. (For example no backticks around the literal values of 1 and False but also you cant mix hyperlinks, such as the attr names, with literal values such as the 1 and False. Therefore writing it in a full sentence is the best way.

stk/calculators/optimization/optimizers.py
```diff
@@ -648,102 +650,503 @@ def optimize(self, mol, conformer=-1):
...
+
+        Parameters
+        ----------
+        output_file : :class:`str`
+            Name of output file with xTB results.
+
+        Returns
+        -------
+        :class:`bool`
+            Returns ``True`` if a negative frequency is present
```
lukasturcani: Missing full stop.

PR #39, opened 2019-06-19
Fix for PDB writer that caused overlap of atom coordinate columns.

andrewtarzia:
Minor changes to _write_pdb function in molecule.py to conform with the PDB column rules


lukasturcani:
I think the current implementation should be fine
the pdb spec reads by column, and the number columns do not need a space between them

andrewtarzia:
I think the issue is actually elsewhere and will update this PR once I have come to the bottom of it.

lukasturcani:
We can reopen this PR once its ready :).

PR #42, opened 2019-06-26
Icosahedron

fiszczyp:
Added construction of icosahedrons as two_plus+five topology.


fiszczyp:
Updated the tests and added one missing new line.

PR #57, opened 2019-09-10
use shell in all cases. modify tests

andrewtarzia:
Fixes issue #49 by setting shell=True in all cases.


lukasturcani:
My xtb tests are failing.

PR #78, opened 2019-10-04
added calculator

lukasturcani:
first

@stevenbennett96

### DIFF
--- a/src/stk/calculators/base_calculators.py
+++ b/src/stk/calculators/base_calculators.py
@@ -1,6 +1,6 @@
 """
-Calculator
-==========
+Base Calculators
+================
 
 
 
@@ -8,9 +8,25 @@ Calculator
 
 
 class Calculator:
-    def __init__(self, use_cache):
+    """
+    A base class for all calculators.
+
+    """
+
+    def __init__(self, **kwargs):
+        return
+
+
+class MoleculeCalculator(Calculator):
+    """
+    Base class for calculators which operate on single molecules.
+
+    """
+
+    def __init__(self, use_cache, **kwargs):
         self._use_cache = use_cache
         self._cache = {}
+        super().__init__(use_cache=use_cache, **kwargs)
 
     def set_cache_use(self, use_cache):
         """
@@ -82,3 +98,12 @@ class Calculator:
         """
 
         return mol in self._cache
+
+
+class PopulationCalculator(Calculator):
+    """
+    Base class for calculators which operate on populations.
+
+    """
+
+    pass
--- a/src/stk/calculators/base_calculators.py
+++ b/src/stk/calculators/base_calculators.py
@@ -39,7 +39,7 @@ class MoleculeCalculator(Calculator):
 
         Returns
         -------
-        :class:`.Calculator`
+        :class:`.MoleculeCalculator`
             The calculator.
 
         """
@@ -73,7 +73,7 @@ class MoleculeCalculator(Calculator):
 
         Returns
         -------
-        :class:`.Calculator`
+        :class:`.MoleculeCalculator`
             The calculator.
 
         """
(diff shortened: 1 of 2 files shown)


src/stk/calculators/calculator.py
```diff
@@ -0,0 +1,84 @@
...
+        Returns
+        -------
+        :class:`.Calculator`
+            The calculator.
+
+        """
+
+        self._use_cache = use_cache
+
+    def is_caching(self):
```
stevenbennett96: This is entirely my own opinion but I think this function name is slightly unclear. I think they should represent an action taken on the class, e.g. ```def check_caching(self):``` .
```is_caching``` sounds more like a variable name because it sounds like a state of something rather than a function.
As a user I would be inclined to use this as 
```python 
calculator.is_caching 
```
rather than 
```python 
calculator.is_caching() 
```

src/stk/calculators/calculator.py
```diff
@@ -0,0 +1,84 @@
...
+        -------
+        :class:`.Calculator`
+            The calculator.
+
+        """
+
+        self._cache.add(mol)
+        return self
+
+    def is_in_cache(self, mol):
```
stevenbennett96: Same as above comment for naming. Maybe name could be ```check_in_cache``` or even ```check_mol_in_cache```, but this may be excessive.

stevenbennett96: Again, this is my own opinion and is subjective.

src/stk/calculators/calculator.py
lukasturcani: The general convention in stk is that a method which returns a bool value begins with is_.  The reason for this is that when you write an if statement it is easy to read. for example

```python
if calculator.is_caching():
    # do stuff
```

So I think this needs to be addressed in a separete github issue - ie file an issue for changing the convention.

src/stk/calculators/base_calculators.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+class MoleculeCalculator(Calculator):
+    """
+    Base class for calculators which operate on single molecules.
+
+    """
+
+    def __init__(self, use_cache, **kwargs):
+        self._use_cache = use_cache
```
stevenbennett96: Why not have `_use_cache` in calculators if it's just a bool. Can we not assume every calculator is going to use the cache?

src/stk/calculators/base_calculators.py
```diff
@@ -0,0 +1,109 @@
...
+            ``True`` if the cache is to be used.
+
+        Returns
+        -------
+        :class:`.Calculator`
+            The calculator.
+
+        """
+
+        self._use_cache = use_cache
```
stevenbennett96: Doesn't return anything

src/stk/calculators/base_calculators.py
```diff
@@ -0,0 +1,109 @@
...
+    def set_cache_use(self, use_cache):
+        """
+        Set cache use on or off.
+
+        Parameters
+        ----------
+        use_cache : :class:`bool`
+            ``True`` if the cache is to be used.
+
+        Returns
```
stevenbennett96: Function returns nothing

src/stk/calculators/base_calculators.py
```diff
@@ -0,0 +1,109 @@
...
+        return mol in self._cache
+
+
+class PopulationCalculator(Calculator):
+    """
+    Base class for calculators which operate on populations.
+
+    """
+
+    pass
```
stevenbennett96: Is this TBC?

src/stk/calculators/base_calculators.py
lukasturcani: Indeed not, for exmple PopulationCalculator has no need for cache - for example a FitnessNormalizer

src/stk/calculators/base_calculators.py
lukasturcani: For now the class does nothing except serve as base class.

PR #79, opened 2019-10-04
Energy calculator

lukasturcani:
after calculator

@stevenbennett96

### DIFF
--- a/src/stk/calculators/energy/energy_calculators.py
+++ b/src/stk/calculators/energy/energy_calculators.py
@@ -80,6 +80,8 @@ from ...utilities import (
     XTBExtractor
 )
 
+from ..calculator import Calculator
+
 
 logger = logging.getLogger(__name__)
 
@@ -128,7 +130,7 @@ def _add_cache_use(get_energy):
     return inner
 
 
-class EnergyCalculator:
+class EnergyCalculator(Calculator):
     """
     Calculates the energy of molecules.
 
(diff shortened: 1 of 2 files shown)


src/stk/calculators/energy/energy_calculators.py
```diff
@@ -128,29 +130,12 @@ def inner(self, mol):
...
 
 
-class EnergyCalculator:
+class EnergyCalculator(Calculator):
     """
     Calculates the energy of molecules.
 
     """
 
-    def __init__(self, use_cache=False):
```
stevenbennett96: This may be a stupid question, but what happens to all the classes that inherit from EnergyCalculator but call 
```python 
super().__init__(some_args)
```
If the __init__ method is removed, do they break?

stevenbennett96: Or is the __init__ inherited from the `Calculator` class?

src/stk/calculators/energy/energy_calculators.py
```diff
@@ -128,29 +130,12 @@ def inner(self, mol):
...
-            calculated value.
-
-        """
-
-        # Maps molecules to previously calculated energy values.
-        self._cache = {}
-        self._use_cache = use_cache
-
     def __init_subclass__(cls, **kwargs):
         cls.get_energy = _add_cache_use(cls.get_energy)
```
stevenbennett96: A comment would be quite useful here explaining that any subclasses' `get_energy` function will be decorated with cache use

src/stk/calculators/energy/energy_calculators.py
lukasturcani: yep, you keep searching up until you hit something, in this case they will hit the init of the Calculator class.

src/stk/calculators/energy/energy_calculators.py
lukasturcani: This is getting addressed in a future PR, in that the entire line of code is being removed.

PR #86, opened 2019-10-10
removed duplication

lukasturcani:
@stevenbennett96

### DIFF
--- a/src/stk/calculators/base_calculators.py
+++ b/src/stk/calculators/base_calculators.py
@@ -52,6 +52,37 @@ class MoleculeCalculator(Calculator):
 
     """
 
+    def _cache_result(self, fn, mol):
+        """
+        Apply `fn` to `mol` and cache the result.
+
+        Parameters
+        ----------
+        fn : :class:`callable`
+            Takes a :class:`.Molecule` as a parameter and returns
+            some result, which is cached, depending on if
+            caching is turned on.
+
+        mol : :class:`.Molecule`
+            The molecule onto which `fn` is applied.
+
+        Returns
+        -------
+        :class:`object`
+            The value returned by ``fn(mol)``.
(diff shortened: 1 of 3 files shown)


src/stk/calculators/base_calculators.py
```diff
@@ -52,6 +52,37 @@ class MoleculeCalculator(Calculator):
 
     """
 
+    def _cache_result(self, fn, mol):
```
stevenbennett96: Could you explain why this is in the abstract base calculator class and not in _MolecularCalculator as an implementation detail?
It looks to me like this function relies on self.is_caching() and self.is_in_cache() being present, however, they are not implemented in the base call.

src/stk/calculators/energy/energy_calculators.py
```diff
@@ -115,15 +115,7 @@ def get_energy(self, mol):
...
 
-        if self.is_caching() and self.is_in_cache(mol):
-            return self.get_cached_value(mol)
-
-        energy = self._get_energy(mol)
-
-        if self.is_caching():
-            self.add_to_cache(mol, energy)
-
-        return energy
```
stevenbennett96: Good removal of code duplication.

src/stk/calculators/base_calculators.py
lukasturcani: Methods in an abstract base class are not expected to be called. This ABC provides provides a customizable algorithm, with the custmization provided by the is_caching(), is_in_cache() calls. A subclass will need to implement the virtual methods and will be allowed to call this algorithm. So specifically, this is ABC does not expect a subclass to implement an algoirithm for _cache_restul. it exepcts sublcassses to use this algorith, but it allows it customization via the virtual method calls.

PR #108, opened 2020-04-29
Modify bb init

andrewtarzia:
Please suggest better ways to do the bonds tuple generation in _init_from_rdkit_mol()


tests/molecular/molecules/building_block/fixtures/init_from_rdkit_mol.py
```diff
@@ -82,6 +86,14 @@
...
             placer_ids=(0, 1, 2, 3),
         ),
+        CaseData(
+            building_block=stk.BuildingBlock.init_from_rdkit_mol(
+                molecule=dative_molecule,
+            ),
+            functional_groups=(),
+            core_atom_ids=(0, 1),
+            placer_ids=(0, 1),
+        ),
```
lukasturcani, line 96: I think the test case for dative bonds needs to go into the molecule case data, because thats that case data that get tested by get_bonds() -> which is what we want to test

src/stk/molecular/molecules/building_block.py
```diff
@@ -553,14 +553,32 @@ def _init_from_rdkit_mol(
...
+                Bond(
+                    atom1=atom1,
+                    atom2=atom2,
+                    order=b.GetBondTypeAsDouble(),
+                    is_dative=is_dative,
+                )
             )
-            for b in molecule.GetBonds()
-        )
+        bonds = tuple(bonds)
```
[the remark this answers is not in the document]
lukasturcani: What if you have a dative bond, going from atom2 to atom1? Won't you need to change the order of that atoms supplied to the Bond initializer?

src/stk/utilities/utilities.py
```diff
@@ -449,6 +450,32 @@ def flatten(iterable, excluded_types=None):
...
+    :class:`bool`
+        ``True`` if atom has the atomic number of a metal atom.
+
+    """
+
+    # Metal atomic numbers.
+    metal_atomic_numbers = set(chain(
+        list(range(21, 31)),
+        list(range(39, 49)),
+        list(range(72, 81))
```
lukasturcani, line 473: Use the ranges without the lists here - the list cause you to waste time and space.

src/stk/molecular/molecules/building_block.py
```diff
@@ -553,14 +552,26 @@ def _init_from_rdkit_mol(
...
         )
-        bonds = tuple(
-            Bond(
-                atom1=atoms[b.GetBeginAtomIdx()],
-                atom2=atoms[b.GetEndAtomIdx()],
+
+        bonds = []
+        for b in molecule.GetBonds():
+            # Check atom ordering.
+            # Atom2 in stk should be the metal atom.
```
lukasturcani, line 559: atom2 not Atom2 I think - since its referring to a variable called atom2, also the comment should explain *why* it should be the metal atom.

src/stk/molecular/molecules/building_block.py
```diff
@@ -553,14 +552,26 @@ def _init_from_rdkit_mol(
...
             for a in molecule.GetAtoms()
         )
-        bonds = tuple(
-            Bond(
-                atom1=atoms[b.GetBeginAtomIdx()],
-                atom2=atoms[b.GetEndAtomIdx()],
+
+        bonds = []
+        for b in molecule.GetBonds():
+            # Check atom ordering.
```
lukasturcani, line 558: This comment is redundant, imo.

src/stk/utilities/utilities.py
```diff
@@ -449,6 +450,32 @@ def flatten(iterable, excluded_types=None):
...
+        An stk Atom.
+
+    Returns
+    -------
+    :class:`bool`
+        ``True`` if atom has the atomic number of a metal atom.
+
+    """
+
+    # Metal atomic numbers.
```
lukasturcani, line 469: This comment is redundant, imo.

src/stk/utilities/utilities.py
```diff
@@ -449,6 +450,32 @@ def flatten(iterable, excluded_types=None):
             yield x
 
 
+def is_metal_atom(atom):
+    """
+    Check if atom has atomic number of a metal atom.
```
lukasturcani, line 455: Reads a bit strangely to me. Also missing markup around atom.
Maybe something like 
```
Check if `atom` is a metal.
``` 
?

src/stk/utilities/utilities.py
```diff
@@ -449,6 +450,32 @@ def flatten(iterable, excluded_types=None):
...
+
+    Parameters
+    ----------
+    atom : :class:`.Atom`
+        An stk Atom.
```
lukasturcani, line 460: no need to write stk here, because the type specification makes it clear that it is an stk atom.

Also if you were referring to stk somewhere in the docs, you would need to add markup.

Atom should also be lowercase.

tests/molecular/molecules/molecule/conftest.py
```diff
@@ -56,6 +63,9 @@
                 num_repeating_units=2,
             ),
         ),
+        stk.BuildingBlock.init_from_rdkit_mol(
+            molecule=dative_molecule,
+        ),
```
lukasturcani, line 68: This is in the wrong place - this fixture does not ge tused by the get_bonds test. You need to add a CaseData to the building_block fixture defined in this file. You will also probably want to move the dative_molecule closer to the building_block fixter in the file.

Also maybe add tests cases for 

`BuildingBlock('dative smiles')`

too

andrewtarzia: `BuildingBlock('dative smiles')` 
will not work because ETKDG will not run on the BuildingBlock initialisation

andrewtarzia: The bond test is also handled in a future PR when changes to to_rdkit_mol is implemented

andrewtarzia: This test actually requires BuildingBlock.to_rdkit_mol so will come in a future PR and is removed for now.

PR #110, opened 2020-04-29
handle is_datives in reactions and tests.

andrewtarzia:
This is implementation to handle the presence of dative bonds in reactions that are not currently defined (functional groups that produce dative bonds will be defined when needed)

### DIFF
--- a/src/stk/molecular/reactions/factories/generic_reaction_factory.py
+++ b/src/stk/molecular/reactions/factories/generic_reaction_factory.py
@@ -26,13 +26,15 @@ def _one_one_reaction(
     functional_group1,
     functional_group2,
     bond_order,
-    periodicity
+    periodicity,
+    is_dative,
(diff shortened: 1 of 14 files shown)


lukasturcani:
I think the current implementation of adding dative bonds has some issues.

`is_datives` currently uses a frozenset to check if a bond should be dative. I think this is a problem because dative bonds in stk have an implied order, but a frozenset does not. Indeed, i think this implementation can easily lead to a metal giving a dative bond to a non-metal - because there is no accounting for ordering the in ReactionFactory.

Also maybe instead of is_datives, dative_bonds would be a better name?

lukasturcani:
I think, keeping the frozenset is fine, but the reaction factory needs to make sure that when it creates a Reaction, the function groups are ordered correctly - such that the dative bond is created correctly.

PR #113, opened 2020-04-29
Smiles key maker

andrewtarzia:


src/stk/molecular/key_makers/smiles.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class Smiles(MoleculeKeyMaker):
+    """
+    Used to get the SMILES of molecules.
+
+    Examples
+    --------
+    You want to use the isomeric, canonical SMILES from RDKit as part
+    of a JSON representation of a molecule
```
lukasturcani, line 18: Each example should have an italicised title of no more than 72 characters, including the newline at the end. For example, `*Adding SMILES to a Molecule's JSON Representation*` - the stars on either side make the title render as italic in the compiled docs

lukasturcani: (I realized, just now, that the other key makers do not have titles for their examples - this is an oversight on my part, look at the examples of topology graphs like Linear Polymer to see what I want)

src/stk/molecular/key_makers/utilities.py
```diff
@@ -37,3 +40,22 @@ def get_inchi_key(molecule):
...
+    ----------
+    molecule : :class:`.Molecule`
+        The molecule whose SMILES is required.
+
+    Returns
+    -------
+    :class:`str`
+        The SMILES.
+
+    """
```
lukasturcani, line 59: empty line should be here

src/stk/molecular/key_makers/utilities.py
```diff
@@ -37,3 +40,22 @@ def get_inchi_key(molecule):
...
+    rdkit_mol = rdkit.RemoveHs(molecule.to_rdkit_mol())
+    return rdkit.MolToSmiles(rdkit_mol)
```
lukasturcani, line 61: I know they are on by default, but it might be a good idea to explicitly set the isomeric and canoncical smiles on, in case there is an api change in rdkit.

tests/databases/constructed_molecule/fixtures/constructed_molecule_mongo_db.py
```diff
@@ -98,6 +98,28 @@
...
+                ),
+            ),
+            key={'SMILES': 'Br[C+2][C+2][C+2][C+2]Br'},
+        ),
+        CaseData(
+            database=stk.ConstructedMoleculeMongoDb(
+                mongo_client=MockMongoClient(),
+                jsonizer=stk.ConstructedMoleculeJsonizer(
+                    key_makers=(stk.Smiles(),)
+                ),
```
lukasturcani, line 122: I don't think we need a new test case for a constructed molecule mongo db - since you didnt change that class.

tests/databases/molecule/fixtures/molecule_mongo_db.py
```diff
@@ -38,6 +38,17 @@
...
+            database=stk.MoleculeMongoDb(
+                mongo_client=MockMongoClient(),
+                lru_cache_size=0,
+                jsonizer=stk.MoleculeJsonizer(
+                    key_makers=(stk.Smiles(),)
+                ),
+            ),
+            molecule=stk.BuildingBlock('BrBr'),
+            key={'SMILES': 'BrBr'},
+        ),
```
lukasturcani, line 51: I don't think we need a new test case for a molecule mongo db, since you didnt change that class.

tests/molecular/key_makers/conftest.py
```diff
@@ -18,6 +18,12 @@
             key_name='InChIKey',
             key='PIICEJLVQHRZGT-UHFFFAOYSA-N',
         ),
+        CaseData(
+            key_maker=stk.Smiles(),
+            molecule=stk.BuildingBlock('NCCN'),
+            key_name='SMILES',
+            key='NCCN',
+        ),
```
lukasturcani, line 26: Maybe a test case to check that the smiles is indeed canonical, and another test case to check that the smiles is isomeric.
FOr the canonical test case if you make a building block like 
`BuildingBlock('C(N)CN')` you should get `'NCCN'` as the smiles also.

src/stk/molecular/key_makers/inchi.py
```diff
@@ -14,6 +14,8 @@ class Inchi(MoleculeKeyMaker):
 
     Examples
     --------
+    *Adding InChi to a Molecule's JSON*
```
lukasturcani, line 17: I at the end of InChI should be capitalized.

src/stk/molecular/key_makers/utilities.py
```diff
@@ -57,5 +57,10 @@ def get_smiles(molecule):
         The SMILES.
 
     """
+
     rdkit_mol = rdkit.RemoveHs(molecule.to_rdkit_mol())
-    return rdkit.MolToSmiles(rdkit_mol)
+    return rdkit.MolToSmiles(
+        rdkit_mol,
```
lukasturcani, line 63: should be `mol=rdkit_mol,`

tests/molecular/key_makers/conftest.py
```diff
@@ -24,6 +24,12 @@
             key_name='SMILES',
             key='NCCN',
         ),
+        CaseData(
+            key_maker=stk.Smiles(),
+            molecule=stk.BuildingBlock('C(N)CN'),
+            key_name='SMILES',
+            key='NCCN',
+        ),
```
lukasturcani, line 32: Any ideas for isomeric smiles test cases?

src/stk/molecular/key_makers/molecule.py
```diff
@@ -106,9 +107,18 @@ def get_key(self, molecule):
...
+            If key of `molecule` is an empty string.
 
-        return self._get_key(molecule)
+        """
+        key = self._get_key(molecule)
+        if not key:
+            raise ValueError(
+                f'Key ({self._key_name}) of {molecule} is empty string'
+            )
+        return key
```
lukasturcani, line 121: Why is this necessary? In addition, the error message is wrong, because the return type of get_key can be any python object, not just a string. 

I also dont think the assumption that an empty string is an invalid key holds in  every possible case - which this code implies. The error raising should be done by the get_key function itself, not by the key maker, imo

src/stk/molecular/key_makers/utilities.py
```diff
@@ -17,7 +17,10 @@ def get_inchi(molecule):
 
     """
 
-    return rdkit.MolToInchi(molecule.to_rdkit_mol())
+    return rdkit.MolToInchi(
+        molecule.to_rdkit_mol(),
```
lukasturcani, line 21: should be `mol=molecule.to_rdkit_mol()`

tests/databases/constructed_molecule/fixtures/constructed_molecule_mongo_db.py
```diff
@@ -98,7 +98,7 @@
                         ),
                     ),
                 ),
-                lru_cache_size=128,
+                lru_cache_size=0,
```
lukasturcani, line 101: why? also - even if I did agree to this, it would have to be a separate PR. I'm guessing this was added here by mistake?

andrewtarzia: I do not even know what that does and did not change it. mustve been a copy and paste error...

PR #115, opened 2020-04-30
Changed docstring a bit.

lukasturcani:
Since valency is not an issue for stk, as it has not concept of it. I think this phrasing is maybe more clear about what a dative bond does?


lukasturcani:
@andrewtarzia

lukasturcani:
Meh, changed my mind, I don't think there's actually a difference between the two statements, its not like stk has a concept of electrons.

PR #123, opened 2020-05-01
Metal centre topology

andrewtarzia:
Implement base classes for metal centre topologies


andrewtarzia:
@lukasturcani if you reject the SingleAtom default to dative bonds in PR #124 , then let me know cause I need to modify this one.

andrewtarzia:
The current commit will fail. need to merge with the branch that has SingleAtom functional group

PR #124, opened 2020-05-01
add single atom reaction factory bond order default

andrewtarzia:


lukasturcani:
Hmm, I'm not sure if we should assume that a Single Atom functional group will be used in dative context most commonly. It seems to be a very generally useful fg to me.  As such, I think there is a a case for keeping the default bond order as 1. ie leaving the defaults unchanged.

andrewtarzia:
I thought you might say that. happy either way to be honest. up to you

lukasturcani:
Aight, let's keep the default as-is for now.

PR #125, opened 2020-05-01
implement metalboundatom FG and factory

andrewtarzia:


andrewtarzia:
this code is not required currently. Closing

PR #137, opened 2020-05-07
Add basic example for metals systems

andrewtarzia:


lukasturcani:
I think the discussion should be moved out of basic_examples and potentially somehwere like metal_complex.py, open to suggestions on this.

I think the discussion should also ignore any discussion of implementatoin details and just act as a straight forward example. If you want to dative bonds, do use bond order 9. There is no need to explain interactions with rdkit. The end user doesnt care, they just want to make dative bonds.

The reasoning behind use of bond order 9 is probably left to some other bits of documetation and some other PR, if necessary.

docs/source/basic_examples.rst
```diff
@@ -555,6 +555,75 @@ other :mod:`~.functional_group_factory`. Note that the number you
...
+
+    # Or using stk.
+    rdkit_mol = bb.to_rdkit_mol()
+
+Dative bonds are not defined for the InChI or InChiIKey functions of
+:mod:`rdkit`. Therefore, when storing metal-containing molecules in a
+database, a different key is required. Because dative bonds are
+implemented in the definition of SMILES, the SMILES string makes a
+useful key for metal-containing molecules. We have implemented a SMILES
+:class:`.KeyMaker` for this purpose
```
lukasturcani, line 605: This can be a its own, separate example.

andrewtarzia:
So, I think I am going to suggest it stays in basic_examples. but much simplified and the metal_complex.py example can go into more detail!

PR #189, opened 2020-07-02
Deprecate ValueMongoDb arguments.

lukasturcani:
This PR replaces the single `lru_cache_size` argument with two separate `cach_size` arguments. This has the benefit of allowing the user to toggle each cache individually. For example, you might want to allow caching for the `put` method, while preventing caching for the `get` method. You might want to prevent caching in the `get` method exclusively in cases where you are adding the same molecule to the database, but in each case it has a different position matrix. In such a situation, the `get` method would not reflect the updates to the database if caching is turned on.

This PR also makes the tests use an actual MongoDb for testing - because it is easier to test than using a Mock database.

### DIFF
--- a/src/stk/databases/mongo_db/value.py
+++ b/src/stk/databases/mongo_db/value.py
@@ -5,6 +5,7 @@ Value MongoDB
 """
 
 from functools import lru_cache
+import warnings
 
 from stk.molecular import InchiKey
 from .utilities import HashableDict
@@ -73,7 +74,9 @@ class ValueMongoDb(ValueDatabase):
         collection,
         database='stk',
         key_makers=(InchiKey(), ),
-        lru_cache_size=128,
+        lru_cache_size='',
+        put_lru_cache_size=128,
+        get_lru_cache_size=128,
         indices=('InChIKey', ),
     ):
         """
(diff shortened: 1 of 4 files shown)


src/stk/databases/mongo_db/value.py
```diff
@@ -98,21 +101,49 @@ def __init__(
             key, they will return the same value from the database.
 
         lru_cache_size : :class:`int`, optional
```
stevenbennett96, line 103: I like that this is being depreciated, rather than remove outwrite.

tests/databases/value/fixtures/mongo_db.py
```diff
@@ -1,26 +1,30 @@
...
 
 @pytest.fixture(
     params=(
         CaseData(
             database=stk.ValueMongoDb(
-                mongo_client=MockMongoClient(),
+                mongo_client=pymongo.MongoClient(),
                 collection='values',
-                lru_cache_size=0,
+                database='_test_database',
```
stevenbennett96, line 14: Out of interest, why use the underscore here?

lukasturcani: Just to make clear its not for regular use & make it unlikely that the test database will conflict with an actual legitimate database in-use by a user. A random string might be in order for that too, but I figured this was fine for now.

src/stk/databases/mongo_db/value.py
lukasturcani: Yep, my goal is to have a relatively stable API from now. So any API change affecting users will ideally have at least a 6 month warning.

src/stk/databases/mongo_db/value.py
```diff
@@ -98,21 +101,49 @@ def __init__(
...
+            Use the `put_lru_cache_size` and `get_lru_cache_size`
+            arguments instead.
+
             A RAM-based least recently used cache is used to avoid
             reading and writing to the database repeatedly. This sets
             the number of values which fit into the LRU cache. If
             ``None``, the cache size will be unlimited.
 
+        put_lru_cache_size : :class:`int`, optional
+            A RAM-based least recently used cache is used to avoid
```
andrewtarzia, line 115: I do not understand 'A RAM-based least'. What does least mean here?

lukasturcani: it should be read as 'A RAM-based least-recently-used-cache'

PR #192, opened 2020-07-02

lukasturcani:
At the moment, calling `put()` would add a new record into the database.  For example, running

```python

db = stk.ValueMongoDb(...)
db.put(molecule, 1)
db.put(molecule, 2)
```

would mean that the MongoDB collection has 2 entries, one which says the molecule has a value of  `1` and one which says the molecule has a value of `2`.

The desired behavior is to find any records in the database which have matching `MoleculeKey` values, and update those records, if they exist, instead of adding new ones. New ones should only be added if no matching record already exists, hence the use of `upsert=True`. Therefore, running the above code should mean that there is only 1 entry in the MongoDB collection.

There is a test for checking that the database was successfully updated, rather than having a new entry added, coming in a later PR.


src/stk/databases/mongo_db/value.py
```diff
@@ -168,7 +168,20 @@ def put(self, molecule, value):
         return self._put(HashableDict(json))
 
     def _put(self, json):
-        self._values.insert_one(json)
+        keys = dict(json)
+        keys.pop('v')
```
stevenbennett96, line 172: What is 'v'?

lukasturcani: This comes from JSON entry actually deposited into the MongoDB. The entry deposited into the MongoDB has the form 

```python

{
    "v": 12
     "KeyName1": "KeyValue1"
    "KeyName2": "KeyValue2"
   # On so on.
}
```

So v is just the value stored in the ValueDB for that molecule. I use `v` instead of `value` to keep the entry as small as possible and therefore keep the MonoDB size as small as possible too.

src/stk/databases/mongo_db/value.py
```diff
@@ -168,7 +168,20 @@ def put(self, molecule, value):
...
+
+        query = {'$or': []}
+        for key, value in keys.items():
+            query['$or'].append({key: value})
+
+        self._values.update_many(
```
stevenbennett96, line 178: Just to check my understanding of pymongo, it will try to find database entries match any of the keys in the form {key: value}, then update the value with `json`. Will this update all the keys it matches in the filter? Or just a single one?

lukasturcani: The intended  behavior is as you described, and it should update all of them. So if you create an entry of the form 

```python

{    'v': 12
    'SMILES': 'C',
    'InChIKey': 'something',
}
```

And the database currently has two entries

```python

{ 'v': 10,
 'SMILES': 'C'
}

```

and 

```python

{ 'v': 21, 
'InChIKey': 'something'
}
```

You should end up with a database that has two entries, both of which are identical and both of which are 

```python
{    'v': 12
    'SMILES': 'C',
    'InChIKey': 'something',
}
```

lukasturcani: I think this is the most reasonable behavior and what most users would expect?

src/stk/databases/mongo_db/value.py
andrewtarzia: This should have a comment to explain what 'v' is.

andrewtarzia:
Unsure if my comment went through, but these put methods should have comments that explain the arbitrary keys used (i.e. 'v')

lukasturcani:
> Unsure if my comment went through, but these put methods should have comments that explain the arbitrary keys used (i.e. 'v')

#209

PR #193, opened 2020-07-02

lukasturcani:
See the previous PR, this solves the same issue but for `MoleculeMongoDb`.


src/stk/databases/mongo_db/molecule.py
```diff
@@ -296,20 +296,27 @@ def put(self, molecule):
...
-        # insert_one(dict(json['molecule'])), which means that a copy
-        # is modified by insert_one and the json in the cache is
-        # not changed.
-
-        self._molecules.insert_one(dict(json['molecule']))
-        self._position_matrices.insert_one(dict(json['matrix']))
+        keys = dict(json['matrix'])
+        keys.pop('m')
+
+        query = {'$or': []}
```
stevenbennett96, line 302: Same comment as from PR #191, will `$or` ensure that all the keys are updated with the new value?

src/stk/databases/mongo_db/molecule.py
```diff
@@ -296,20 +296,27 @@ def put(self, molecule):
...
+        for key, value in keys.items():
+            query['$or'].append({key: value})
+
+        self._molecules.update_many(
```
stevenbennett96, line 306: I assume that both the `molecules` and `position_matrices` databases are using identical molecule keys?

src/stk/databases/mongo_db/molecule.py
```diff
@@ -296,20 +296,27 @@ def put(self, molecule):
...
+            filter=query,
+            update={
+                '$set': json['molecule'],
```
stevenbennett96, line 309: If I wanted to build a molecule from the database that I called `put` on, would information be required from both `moleucles` and `position_matrices`, or could I just use `molecules`?
I'm a little confused about what `json['molecules]` contains.

src/stk/databases/mongo_db/molecule.py
```diff
@@ -296,20 +296,27 @@ def put(self, molecule):
...
-        # because the one in the cache has the extra items added by
-        # insert_one(). To prevent this use
-        # insert_one(dict(json['molecule'])), which means that a copy
-        # is modified by insert_one and the json in the cache is
-        # not changed.
-
-        self._molecules.insert_one(dict(json['molecule']))
-        self._position_matrices.insert_one(dict(json['matrix']))
+        keys = dict(json['matrix'])
+        keys.pop('m')
```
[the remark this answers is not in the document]
lukasturcani: Nope, `m` holds the position matrix to be deposited into the MongoDB.

Each molecule creates two separate entries which are deposited into two separate collections in the MongoDB.

The first entry which contains information about atoms and bonds is held by `json['moleucle']` and this is deposited into `self._molecules`

The second entry contains the position matix and is accessed by `json['matrix']` and this is deposited into `self._position_matrices`.

src/stk/databases/mongo_db/molecule.py
lukasturcani: Same answer, yes, that is the intended behavior.

src/stk/databases/mongo_db/molecule.py
lukasturcani: Correct! That's how you are able to reconstruct your molecule when you supply a given key.

src/stk/databases/mongo_db/molecule.py
lukasturcani: `json['molecues']` is the entry holding atom and bond information. `json['matrix']` is the entry holding the position matrix. You need both to reconstruct an `stk` `Molecule`.

src/stk/databases/mongo_db/molecule.py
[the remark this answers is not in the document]
andrewtarzia: This should have a comment to explain what m is.

andrewtarzia:
Unsure if my comment went through, but these put methods should have comments that explain the arbitrary keys used (i.e. 'v')

PR #194, opened 2020-07-02

lukasturcani:
This PR does two things. First it does the same thing as the previous PR but for `ConstrcutedMoleculeMongoDb`.

Second it introduces a separate collection for position matrices of building blocks in the MongoDB database. This is necessary because usually a collection of position matrices holds position matrices acquired through some specific method. For example, you would have a collection of position matrices collected from XTB and another collection for position matrices collected from macromodel.

```python

xtb_db = stk.ConstructedMoleculeMongoDb(
    position_matrix_collection='xtb_matrices',
)
macromodel_db = stk.ConstructedMoleculeMongoDb(
    position_matrix_collection='macromodel_matrices',
)

molecule = stk.ConstructedMolecule(...)
xtb_molecule = molecule.with_position_matrix(
    position_matrix=get_xtb_position_matrix(molecule),
)
macromodel_molecule = molecule.with_position_matrix(
    position_matrix=get_macromodel_position_matrix(molecule),
)

xtb_db.put(xtb_molecule)
macromodel_db.put(macromodel_molecule)

```

Note the issue in the last two lines of code 

```python

# Currently, the position matrices of building blocks of xtb_molecule
# are also deposited into the xtb_matrices collection.
xtb_db.put(xtb_molecule)
# Currently the position matrices of building blocks of macromodel_molecule
# are also deposited into the macromodel_matrices collection.
macromodel_db.put(macromodel_molecule)

# This makes no sense because the building blocks of xtb_molecule and
# macromodel_molecule are the exact same BuildingBlock instances, which
# may not have been optimized with either of the two mehods!

```

Therefore a separate argument is needed, which allows the user to specify where the position matrices of building blocks should be deposited.

```python

db = stk.ConstructedMoleculeMongoDb(
    position_matrix_collection='xtb_matrices',
    building_block_position_matrix_collection='etkdg_matrices'
)

```


src/stk/databases/mongo_db/constructed_molecule.py
```diff
@@ -334,32 +355,130 @@ def make_hashable(json):
...
+            molecule_entry['BB']
+            for molecule_entry
+            in self._constructed_molecules.find(query)
+        )
+        for entry_building_block_keys in database_building_block_keys:
+            for keys1, keys2 in zip(
+                building_block_keys,
+                entry_building_block_keys,
+            ):
+                keys1.update(keys2)
```
stevenbennett96, line 480: Just to check, when you iterate over the keys, are actual keys in building_block_keys modified rather than creating a copy as you iterate over it?

src/stk/databases/mongo_db/constructed_molecule.py
```diff
@@ -281,6 +289,9 @@ def __init__(
             constructed_molecule_collection
         ]
         self._position_matrices = database[position_matrix_collection]
+        self._building_block_position_matrices = database[
```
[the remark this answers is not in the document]
lukasturcani: Mostly correct. The `ConstructedMolecule` holds a `Molecule`, not a `BuildingBlock`, but yes.

src/stk/databases/mongo_db/constructed_molecule.py
lukasturcani: No copy is made. The `dicts` in `building_block_keys` are updated with any missing keys, which are retrieved from the MongoDB.

src/stk/databases/mongo_db/constructed_molecule.py
```diff
@@ -331,32 +352,130 @@ def make_hashable(json):
...
+        Add previously deposited keys to `building_block_keys`.
+
+        Checks the constructed molecule collection to find all
+        constructed molecule entries which match `query`. All matches
+        should merely be duplicate entries for the same constructed
+        molecule.
+
+        Each entry for the constructed molecule will have a
+        :class:`list` of building blocks, which were used to
+        construct the constructed molecule.
```
andrewtarzia, line 426: I am assuming their defined positions in the constructed molecule are also held somewhere (i.e. the topology graph)? Otherwise, this is required because you cannot reliably recreate all constructions without knowing all of this information.

lukasturcani: The topology graph is not a part of the ConstructedMolecule -> it is not deposited in to the database as is not needed to restore a ConstructedMolecule.

PR #196, opened 2020-07-02
Added a test to make sure caching is working

lukasturcani:
Added a test to make sure that the `put()` and `get()` caches are getting hit when those methods are called repeatedly.


tests/databases/value/mongo_db/test_value_caching.py
```diff
@@ -0,0 +1,48 @@
...
+        database=database_name,
+    )
+    molecule = stk.BuildingBlock('CCC')
+    database.put(molecule, 43)
+    database.get(molecule)
+    database.get(molecule)
+
+    cache_info = database._get.cache_info()
+    assert cache_info.hits == 1
+    assert cache_info.misses == 1
```
stevenbennett96, line 48: Maybe it's also good to test the cache size in both `put` and `get`? Asserting maxsize value is what we think it is could be useful to check all class arguments are working.

lukasturcani: The cache size is tested for both `put` and `get` but across different tests. I think that makes more sense as they are different methods and should therefore be tested separately IMO.

I think that somewhat conflates the purpose of the test. The test is meant to purely to test whether caching is taking place. To test if arguments are working should be done in a separate test IMO.

PR #197, opened 2020-07-02
Added a test to make sure caching is working.

lukasturcani:


tests/databases/molecule/mongo_db/test_molecule_caching.py
```diff
@@ -0,0 +1,55 @@
...
+    database_name = '_test_get_caching'
+    client = pymongo.MongoClient()
+    client.drop_database(database_name)
+
+    database = stk.MoleculeMongoDb(
+        mongo_client=client,
+        database=database_name,
+    )
+    molecule = stk.BuildingBlock('CCC')
+    database.put(molecule)
```
stevenbennett96, line 43: MAybe add the InChIKey JSON keymarker you're using here as an argument to `database.put()`. It would make it easier to understand why you're using `stk.InChIKey()` in the following `database.get` call.

PR #199, opened 2020-07-02
Added utilities for tracking MongoDB state.

lukasturcani:


tests/databases/utilities.py
```diff
@@ -0,0 +1,130 @@
+def assert_database_state(state1, state2):
```
stevenbennett96, line 1: I'm confused what this is adding? Is it just so it's clearer what you're doing when you compare two databases?

tests/databases/utilities.py
```diff
@@ -0,0 +1,130 @@
...
+
+    return DatabaseEntry(**{
+        key: value
+        for key, value in to_hashable(item.items()) if key != '_id'
+    })
+
+
+class DatabaseEntry:
+    """
+    Represents an record in a MongoDB collection.
```
stevenbennett96, line 70: a record?

tests/databases/utilities.py
lukasturcani: Pretty much, it also means that in cases where errors happen you get additional info, because state1 and state2 get printed.

tests/databases/utilities.py
```diff
@@ -0,0 +1,130 @@
+def assert_database_state(state1, state2):
+    assert state1 == state2
+
+
+class HashableDict(dict):
```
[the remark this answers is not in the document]
lukasturcani: Indeed, it allows the use of dictionaries as dictionary keys.

tests/databases/utilities.py
lukasturcani: Synonym for entry. Will use entry for consistency.

tests/databases/utilities.py
```diff
@@ -0,0 +1,130 @@
...
+    def __hash__(self):
+        return hash((frozenset(self), frozenset(self.values())))
+
+    def __eq__(self, other):
+        return super().__eq__(other)
+
+
+def to_hashable(items):
+    for key, value in items:
+        if key == 'a':
```
andrewtarzia, line 15: These arbitrary key names need comments.

lukasturcani: I take your point, but I think its important to note that they are not arbitrary, but they are short versions of the actual names. 'a' = 'atoms', 'b' = 'bonds, 'BB' = building blocks, etc.

andrewtarzia: Obviously, yes, but its still not as clear as it could be with some comments.

PR #200, opened 2020-07-02
Added a test to make sure database is updating.

lukasturcani:


tests/databases/value/mongo_db/test_value_update.py
```diff
@@ -0,0 +1,233 @@
...
+    database.put(molecule, 43)
+    assert_database_state(
+        state1=get_database_state(database),
+        state2=DatabaseState({
+            DatabaseEntry(
+                InChIKey=stk.InchiKey().get_key(molecule),
+                v=43,
+            ): 1,
+        }),
+    )
```
stevenbennett96, line 55: Aha so the DatabaseState and DatabaseEntry methods are primarily used for testing?

stevenbennett96: I assume that when assert_database_state is called, all of the other information stored in the database that are associated with that molecule are also asserted equal? Such as the atom and bond information

lukasturcani: Indeed, it asserts that the database has the exact molecules, atoms bonds, position matrices etc. It asserts the entire database state is exactly equal to what I expect.

lukasturcani: However - note that the ValueDatabase does NOT store molecules, atoms, bonds or position matrices. So in this case what you said is false. Because this information is not in the database  - nor is it expected to be.

tests/databases/value/mongo_db/test_value_update.py
```diff
@@ -0,0 +1,233 @@
...
+
+
+def test_update_2():
+    """
+    Test that existing entries are updated.
+
+    In this test, your first create two separate entries, using
```
andrewtarzia, line 62: 'your' to 'you'

tests/databases/value/mongo_db/test_value_update.py
```diff
@@ -0,0 +1,233 @@
...
+                InChIKey=stk.InchiKey().get_key(molecule),
+                v=12,
+            ): 1,
+        }),
+    )
+
+    # Should add another entry, as a different key maker is used.
+    database2.put(molecule, 32)
+    assert_database_state(
+        state1=get_database_state(database1),
```
andrewtarzia, line 121: should this not be database2? If not, then I am confused what this test is doing and the docs should be updated.

tests/databases/value/mongo_db/test_value_update.py
```diff
@@ -0,0 +1,233 @@
...
+                SMILES=stk.Smiles().get_key(molecule),
+                v=32,
+            ): 1,
+        }),
+    )
+
+    # Should update both entries as both key makers are used.
+    database3.put(molecule, 56)
+    assert_database_state(
+        state1=get_database_state(database1),
```
andrewtarzia, line 137: should this not be database3? If not, then I am confused what this test is doing and the docs should be updated.

tests/databases/value/mongo_db/test_value_update.py
```diff
@@ -0,0 +1,233 @@
...
+            ): 2,
+        }),
+    )
+
+
+def test_update_3():
+    """
+    Test that existing entries are updated.
+
+    In this test, your first create one entry with two keys. Then
```
andrewtarzia, line 152: 'you'

tests/databases/value/mongo_db/test_value_update.py
andrewtarzia: I think I get it now, they are the same database. But your docs should be more clear.

lukasturcani: Any more specific suggestions?

tests/databases/value/mongo_db/test_value_update.py
lukasturcani: The MongoDb databases and collections accessed by `database1 ` `database2` and  `database3` are all the same. The only difference is that when you use 1, 2 or 3, `.put()`, the entries hold different MoleculeKeys

lukasturcani: This test make sure that when calling `put()` and dpositing different keys, the entries get updated appropriately rather than making new entries or losing old MoleculeKeys

PR #202, opened 2020-07-02
Added a test to make sure database is updating.

lukasturcani:


tests/databases/constructed_molecule/mongo_db/test_constructed_molecule_update.py
```diff
@@ -0,0 +1,426 @@
...
+    json = dict(json)
+    return {
+        'm': tuple(tuple(row) for row in json.pop('m')),
+        **json,
+    }
+
+
+def to_hashable_constructed_molecule(json):
+    json = dict(json)
+    return {
```
andrewtarzia, line 24: arbitrary dict keys should be explained.

tests/databases/constructed_molecule/mongo_db/test_constructed_molecule_update.py
```diff
@@ -0,0 +1,426 @@
...
+    )
+
+    molecule = stk.BuildingBlock(
+        smiles='BrCCCBr',
+        functional_groups=[stk.BromoFactory()],
+    ).with_canonical_atom_ordering()
+
+    polymer1 = stk.ConstructedMolecule(
+        topology_graph=stk.polymer.Linear(
+            # Use it as a building block twice, to make sure it is
```
andrewtarzia, line 174: This does not make sense - why do you need to do it and is this something the user needs to be aware of?

lukasturcani: It is purely for testing purposes. If I deposit the polymer, which has a building block twice, I want to make sure it does not create two entries for that building block but just 1.

PR #212, opened 2020-07-17
Fix bug in example.

andrewtarzia:


lukasturcani:
Where's the bug?

andrewtarzia:
It was a few lines before those that were changed. there was a usage of 'guest' when it was not assigned in the attributes!

lukasturcani:
Why does it need to be assigned to an attribute? I find binding it directly to the function parameter is more clear. You do not have to search for what the attribute holds, the line itself tells you what exactly `guest` is, without scanning for other lines.

lukasturcani:
Update for future reference: In the example that was changed, `guest.get_direction()` was called, which means `guest` needs to be defined. The `guest.get_direction()` call is on line 58 (before change) / 59 (after change).

PR #220, opened 2020-08-04
Get entries api

### LINKED ISSUE #186: New feature: database.get_molecules()/get_entries()
Add simple API for getting an iterable of all entries in a database.

Potentially handle any duplicates (an optional flag to the function call)

andrewtarzia:
Resolves #186

### DIFF
--- a/src/stk/databases/value.py
+++ b/src/stk/databases/value.py
@@ -22,6 +22,25 @@ class ValueDatabase:
     :mod:`molecule_value_database <.databases.value>`,
     can serve as good examples.
 
+    *Iterating Through Entries in Database*
+
+    The :meth:`.get_entries` method of a molecule database instance can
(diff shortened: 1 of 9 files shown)


src/stk/databases/value.py
```diff
@@ -22,6 +22,29 @@ class ValueDatabase:
...
+
+        client = pymongo.MongoClient()
+        molecule_db = stk.MoleculeMongoDb(client)
+
+        value_db = stk.ValueMongoDb(
+            mongo_client=client,
+            collection='atom_counts',
+        )
+
+        for key, molecule in molecule_db.get_all():
```
lukasturcani, line 41: `get_all()` does not yield the key as currently implemented. Plus there are multiple possible keys.

andrewtarzia: Updated.

Issue #219, opened 2020-08-04
Add get_deleted_bonds() to ReactionResult interface.


lukasturcani:
Also make sure construction makes use of this new interface.

@andrewtarzia

PR #221, opened 2020-08-04
Add non-implemented get_deleted_bonds interface.

andrewtarzia:
Adds interface requested in issue #219 

Currently, no reactions actually use this.


src/stk/molecular/topology_graphs/topology_graph/construction_state/molecule_state/deletions_summary.py
```diff
@@ -57,6 +59,9 @@ def __init__(
         deleted_ids : :class:`iterable` of :class:`int`
             The ids of `atoms`, which should be deleted.
 
+        deleted_bond_ids : :class:`iterable` of :class:`tuple`
+            Tuple of atom ids of bonds that should be deleted.
```
lukasturcani, line 63: Might be worth creating a namedtuple and the using it here, so that the shape of the iterable is easy to understand.


(The docstring doesn't include the type because the code has a type hint.)


```python
from typing import NamedTuple

class _BondId(NamedTuple):
    """
    Identifies a bond in a molecule.

    Attributes
    ----------
    atom1_id
        The id of the first atom in the bond.
    
    atom2_id 
        The id of the sceond atom in the bond.

    """

    atom1_id: int
    atom2_id: int

```

```
    deleted_bond_ids : :class:`iterable` of :class:`._BondId`
        Ids of bonds that should be deleted
```

lukasturcani (changes requested):
One other thing is that it would be nice to include a reaction which does delete bonds, so that we have a test case for that. Can be super dumb / unrealistic.

andrewtarzia:
I will also add a reaction that does delete bonds to test this.

src/stk/molecular/topology_graphs/topology_graph/construction_state/molecule_state/deletions_summary.py
```diff
@@ -33,6 +34,7 @@ def __init__(
         bond_infos,
         position_matrix,
         deleted_ids,
```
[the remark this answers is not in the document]
andrewtarzia: Done

src/stk/molecular/topology_graphs/topology_graph/construction_state/molecule_state/deletions_summary.py
andrewtarzia: Done. Added it to reactions_summary.py and updated docs.

andrewtarzia:
I propose this is not so straight forward and requires many more changes:

* an interface for the RingAmineReaction is required - currently it cannot be used through a Factory.
* A BreakBondReactionFactory is required - this will be used in-place of a GenericReactionFactory
* FunctionalGroup API needs updating to be able to define an FG that can lose bonds (currently deleters and bonders cannot define this without breaking their meaning) - this is not actually true, just need a new FunctionalGroup class, akin to GenericFunctionalGroup, which handles this API ->
* a Functional Group (eg. `BreakBond` and `BreakBondFactory`), which take in a Smarts string to define a sub structure and bond-to-be-broken atom ids (whatever we want to call them)

Therefore, I suggest that adding a use of this and test will be a later PR alongside these changes.

lukasturcani (approved):
Ty! LGTM.

Issue #227, opened 2020-08-20
.with_structure_from_file methods should handle formatting changes for periodic file types


andrewtarzia:
Only .pdb and .coord file types are relevant. Unsure what the impact is at this point, but the change in file formatting may break the current code if it expects a rigid formatting.

lukasturcani:
Do we have a way forward with this for .pdb files?

andrewtarzia:
We do not have a with_structure_from_file for PDB files because we use the RDKIT function for PDB files in BuildingBlock.

The solution would actually just be to add a with_structure_from_file for pdb that uses the rdkit code - thoughts?

lukasturcani:
Yes, looks like we use the rdkit pdb reader for `BuildingBlock.init_from_file`. We also have our own, custom, pdb writer for `Molecule.write`. Remind me, what exactly do we want to be able to do? Just read the file, and make sure the function does not crash in cases where periodic parametes are? Do we want to read the periodic parameters? What would the periodic parameters be? What is generating these periodic files we are trying to read?

andrewtarzia:
-  Remind me, what exactly do we want to be able to do? **Read a PDB without crashing if the file is not fitting a rigid format (not relevant for pdb because we do not handle it)**
- Just read the file, and make sure the function does not crash in cases where periodic parametes are defined in the file? **Indeed**.
- Do we want to read the periodic parameters? What would the periodic parameters be? What is generating these periodic files we are trying to read?: **At this stage, no, stk does not update cell parameters when updating a structure - I do not see it needing to do this. stko does though. Optimisations generate new unit cells.**

lukasturcani:
Ok cool. So it does seem like just adding the rdkit reader to the update function should do the trick. But if this is so, I wonder why pdb is not supported already? Was there an issue, or was it just an innocent omission?

I guess there's only one way to find out, add it and see if the tests pass. Are you happy to do it? I could try if you're busy, but I'm not sure how to go about getting good test cases.

andrewtarzia:
Ya, I will have a crack this afternoon. I think the omission was accidental, because general use-cases are always .mol files

PR #239, opened 2020-08-28
Update turbomole updater class

andrewtarzia:
Updates turbomole updater (half sorts issue #227) to:

-  check the coordinate units (although fractional coordinates not implemented because complex, and unnecessary - an error is implemented if this comes up)

- coordinate section placement in file is not hardcoded.


src/stk/molecular/molecules/molecule/utilities/updaters/turbomole.py
```diff
@@ -25,16 +26,39 @@ def _with_structure_from_turbomole(self, path):
...
-        _, *content, __ = f.readlines()
+        content = f.readlines()
+
+    for i, cont in enumerate(content):
+        if '$coord' in cont:
+            if 'angs' in cont:
+                coord_units = 'angstrom'
+            elif 'frac' in cont:
+                coord_units = 'fractional'
+                raise NotImplementedError(
```
lukasturcani, line 46: Based on https://docs.python.org/3/library/exceptions.html#NotImplementedError I'm not sure this is the right error type. Happy to be convinced otherwise.

src/stk/molecular/molecules/molecule/utilities/updaters/turbomole.py
```diff
@@ -25,16 +26,39 @@ def _with_structure_from_turbomole(self, path):
...
+
     """
 
     bohr_to_ang = 0.5291772105638411
 
     with open(path, 'r') as f:
-        _, *content, __ = f.readlines()
+        content = f.readlines()
+
+    for i, cont in enumerate(content):
```
lukasturcani, line 40: `cont` is not a good name for a variable imo. `line` may be better here?

lukasturcani, line 40: Also not a huge for of `i`, I know its a common convention, but `line_number` I think is more clear, especailly because later you define `j` too at which point I think the code becomes ... coded.

src/stk/molecular/molecules/molecule/utilities/updaters/turbomole.py
```diff
@@ -25,16 +26,39 @@ def _with_structure_from_turbomole(self, path):
...
+                coord_units = 'fractional'
+                raise NotImplementedError(
+                    'Fractional coordinates are not handled currently.'
+                )
+            elif 'bohr' in cont:
+                coord_units = 'bohr'
+            else:
+                coord_units = 'bohr'
+            coord_section = []
+            for j in content[i+1:]:
```
lukasturcani, line 54: `subsequent_line` instead of `j` maybe?

andrewtarzia:
This actually handles all of issue #227 because there is no PDB updater.

src/stk/molecular/molecules/molecule/utilities/updaters/turbomole.py
lukasturcani, line 40: Actually, is this loop necessary? Why not re-use the old code

```python

header, *content, _ = f.readlines()
if 'frac' in header:
    raise RuntimeError('File contains fractional coordinates, which are not currently supported.')
```

lukasturcani (changes requested):

src/stk/molecular/molecules/molecule/utilities/updaters/turbomole.py
andrewtarzia: Done

src/stk/molecular/molecules/molecule/utilities/updaters/turbomole.py
andrewtarzia: Done.

andrewtarzia: Done.

src/stk/molecular/molecules/molecule/utilities/updaters/turbomole.py
andrewtarzia: Done.

src/stk/molecular/molecules/molecule/utilities/updaters/turbomole.py
andrewtarzia: Explained on Slack.

lukasturcani (approved):

PR #244, opened 2020-09-22
Adding 2 factories to allow Disulfide and Alkene exchange reactions

zaeemnajeeb:
Addition of 2 new Function Group Factories (FGF). these are the Alkene Metathesis Factory and the Disulfide Exchange Factory. They create `.Alkene` and `.Disulfide` instances so no new FGs needed to be added.

Works similar to a FGF except has a new variable upon initialisation which is `largest_chain` which is set to True by default. This will assign the bonders and deleters according to what would delete the fewest atoms. This is done by iterating over the molecule's connected atoms from the stated deleter atom using the bonds. This is done via the new utility function, `_find_R_ids`.

If a user needs to explicitly state the bonders, they need to set `largest_chain` to false, and then state the bonders in the usual format, except each alkene will correspond to a single integer in bonders. Therefore, a molecule with 5 alkenes in it would require bonders to have 5 integers in a tuple, ordered according to the order the alkenes are assigned (based on atom_ids). These integers will be either 0 or 3, corresponding to the functional group string in the docs. Examples are provided in the doc string.

### DIFF
--- a/src/stk/databases/value.py
+++ b/src/stk/databases/value.py
@@ -24,7 +24,7 @@ class ValueDatabase:
 
     *Iterating Through Entries in Database*
 
-    The :meth:`.get_entries` method of a molecule database instance can
+    The :meth:`.get_all` method of a molecule database instance can
     be used to iterate through entries, the keys of which can be used
     to access values.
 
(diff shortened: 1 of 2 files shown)


zaeemnajeeb:
This pull request was made in the wrong place

Issue #184, opened 2020-06-29
Cumulative error in rotations


andrewtarzia:
Multiplication of rotation matrices upon construction cause a decrease in building block diameter. We expect this to be due to numerical errors in the multiplication, which accumulate over time.

Normalizing matrices is the first step in solving. But a change to using quaternions and to doing all rotations in one step.

Minimum working example:
[cubes.zip](https://github.com/lukasturcani/stk/files/4845791/cubes.zip)

andrewtarzia:
How come this is not in the stk paper project?

lukasturcani:
I can add it, but I did not consider this priority. In that I don't think this is a big problem atm (might be wrong here), and it does not actually affect the content of the paper.

andrewtarzia:
I think we should add it, and I do think it is important to remove bugs like this before publication.

lukasturcani:
I cannot reproduce the error from the example. Both of the numbers printed are identical.

```
▶ python min_example.py
init  16.246521105147405
const 16.246521105147405
```

andrewtarzia:
I also cannot reproduce this with the latest version of stk (master) - will explore further.

lukasturcani:
Will re-open if reproducible.

andrewtarzia:
This issue is definitely still occurring, see the attached mol file (as a text file): look at the nodes on the corner of the cubes, one is much smaller than the rest. Will need to figure out how to reproduce on small scale. We must have changed something about the building blocks ability to update that messed with the minimum example 
[TEMPdel.txt](https://github.com/lukasturcani/stk/files/5505293/TEMPdel.txt)

lukasturcani:
Right, so maybe the building block is not changing but the problem is that when placed on a topology graph, the building black size is different? That makes much more sense, because building blocks are and were immutable. However, the modified clones may have ended up with cumulative errors?

andrewtarzia:
Yes, I think that is the case. So the issue definitely still needs fixing.

In the old version the example did work.

lukasturcani:
Can you still update the example please with comments that specify exactly what is wrong. Just so that its super for someone who opens it up with no context.

andrewtarzia:
It is hard to state exactly what is wrong in the code example, I updated the ZIP below and will state the atoms to look for to visualise the effect.

The cluster bonded to the Zinc atom with id 632.

[cubes.zip](https://github.com/lukasturcani/stk/files/5509141/cubes.zip)

[...]

PR #255, opened 2020-11-19
Normalising rotation matrices

andrewtarzia:
This is fix for #184. Handles the issue by ensuring all rotation matrices are normalised using scipy Rotation class.

In future (an enhancement), I would think we could use that class to take a rotation matrix, and output a quaternion (not limited to this, but this is my first guess) for stable molecule manipulation.


andrewtarzia:
To solve the issue with this test (1 molecule (metal_cage_m4l6_tetrahedron-metal_cage_m4l6_tetrahedron0) failing the get_plane_normal test for two random atom_id inputs, which passed on multiple local Ubuntu machines), I have changed the molecule that is failing to a different molecule based on the input building blocks.

lukasturcani:
How come the iron complex was replaced with the fake complex?

lukasturcani (approved):

lukasturcani (changes requested):
Can confirm that this fix works, tyvm!. Just one thing: could you add the molecule which caused the issue in https://github.com/lukasturcani/stk/issues/184 as a test case? And when you do, add a little comment above it which says something along the lines of "See https://github.com/lukasturcani/stk/issues/184"

lukasturcani:
Retracting my previous comment because the issue is visible in an existing test case.

Issue #205, opened 2020-07-04
All topology graphs need to specify how many functional groups are required on building blocks.


lukasturcani:
Notably cage topology graphs are missing this information.

lukasturcani:
Mostly addressed by #256 but COF topology graphs still need this information.

lukasturcani:
Finished by https://github.com/lukasturcani/stk/pull/264

PR #256, opened 2020-11-27
Define connectivity

andrewtarzia:
Solves issue #205


lukasturcani:
TY! What do you think about also adding an example? Also, I think that saying "When using a dict for initialization" is potentially confusing, because a dict can be used in two places: for `building_blocks` but also for `vertex_alignments`. So maybe something like

```
When using a :class:`dict` for the `building_blocks` parameter,  as in :ref:`multi-building block cage construction`,
a :class:`.BuildingBlock` needs to be assigned to each for the following vertex ids:

    | 4-functional group: 0 to 5
    | 3-functoinal group: 6 to 13

```

(vertex ids instead of numbers, I think may be more descriptive as to the purpose?)

also note the link I added, so that the sentence links directly to where Multi-parameter example is written. I think that providing the numbers is a little abstract, so linking to the concrete example of dict usage is pretty helpful. (thoughts?)

this will mean that the following will need to be added in ``stk.molecule.topology_graphs.cage.cage``

```
.. _`multi-building block cage construction`:

*Multi-Building Block Cage Construction*
```

so that the :ref: works

andrewtarzia:
I like the idea and logic, I will get on it soon!

andrewtarzia:
Updated. Would be keen to see this in the actual docs for formatting - can you write how the docs are compiled?

lukasturcani:
Yep, I wanted the same, so I've added your branch to my repostitory under `andrewtarzia/define_connectivity`. I then told my read the docs to compile the docs for that branch. (Note that this is from yesterday, I haven't recompiled them yet.)

https://stk.readthedocs.io/en/andrewtarzia-define_connectivity/

That being said, there are two ways to compile the docs - and in fact, the contents of this comment should probably added to the developer's guide when it's written. Oh and it might be good if you check both ways and ensure I didn't miss a step. It was a while since I set up readthedocs, there may be a gotcha I forgot about.

### Locally

This is the simplest and fastest thing to do.

```shell script

# Go into the root of the stk repository.
cd path/to/stk

# Make a new conda environment, which holds dependencies for compiling the docs.
conda create --name stk-docs pip

# Activate the environment.
conda activate stk-docs

# Install the dependencies.
pip install -r docs/requirements.txt

# Extract docstrings into rst files.
cd docs
./remake_modules.bash

# Build the html documentation.
make html

```

you should now have the following folder (where `stk` is the root of the `stk` repository)
`stk/docs/build/html`, which holds all the compiled docs. You can open `/path/to/stk/docs//build/html/index.html` in your browser.

### ReadTheDocs

Umm, lets do a call and try this togther and then I can update this comment? I want to make sure I don't miss any gotchas.

andrewtarzia:
Done!

src/stk/molecular/topology_graphs/cage/metal_topologies/m4l6_tetrahedron.py
```diff
@@ -20,6 +20,11 @@ class M4L6Tetrahedron(Cage):
     Building blocks with three functional groups are required for this
     topology.
 
+    When using a :class:`dict` for the `building_blocks` parameter,
+    as in :ref:`multi-building-block-cage-construction`:
+    *Multi-Building Block Cage Construction*, a :class:`.BuildingBlock`
+    needs to be assigned to each of the following vertex ids:
```
lukasturcani, line 26: This one was missed looks like.

src/stk/molecular/topology_graphs/cage/metal_topologies/m8l6_cube.py
```diff
@@ -19,12 +19,13 @@ class M8L6Cube(Cage):
...
-    following numbers:
+    When using a :class:`dict` for the `building_blocks` parameter,
+    as in :ref:`multi-building-block-cage-construction`:
+    *Multi-Building Block Cage Construction*, a :class:`.BuildingBlock`
+    needs to be assigned to each of the following vertex ids:
 
-        | metals (3 functional groups): 0 to 7
-        | ligands (4 functional groups): 8 to 13
+        | 3-functional group: 0 to 7
+        | 4-functoinal group: 8 to 13
```
lukasturcani, line 28: typo

src/stk/molecular/topology_graphs/cage/metal_topologies/m12l24.py
```diff
@@ -19,12 +19,13 @@ class M12L24(Cage):
...
-    following numbers:
+    When using a :class:`dict` for the `building_blocks` parameter,
+    as in :ref:`multi-building-block-cage-construction`:
+    *Multi-Building Block Cage Construction*, a :class:`.BuildingBlock`
+    needs to be assigned to each of the following vertex ids:
 
-        | metals: 0 to 11
-        | ligands: 12 to 35
+        | 4-functional group: 0 to 11
+        | 2-functoinal group: 12 to 35
```
lukasturcani, line 28: typo

src/stk/molecular/topology_graphs/cage/metal_topologies/m24l48.py
```diff
@@ -19,12 +19,13 @@ class M24L48(Cage):
...
-    following numbers:
+    When using a :class:`dict` for the `building_blocks` parameter,
+    as in :ref:`multi-building-block-cage-construction`:
+    *Multi-Building Block Cage Construction*, a :class:`.BuildingBlock`
+    needs to be assigned to each of the following vertex ids:
 
-        | metals: 0 to 23
-        | ligands: 24 to 71
+        | 4-functional group: 0 to 23
+        | 2-functoinal group: 24 to 71
```
lukasturcani, line 28: typo

src/stk/molecular/topology_graphs/cage/metal_topologies/m6l12_cube.py
```diff
@@ -21,12 +21,13 @@ class M6L12Cube(Cage):
...
-    following numbers:
+    When using a :class:`dict` for the `building_blocks` parameter,
+    as in :ref:`multi-building-block-cage-construction`:
+    *Multi-Building Block Cage Construction*, a :class:`.BuildingBlock`
+    needs to be assigned to each of the following vertex ids:
 
-        | metals: (0, 1, 2, 3, 4, 5)
-        | ligands: (6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17)
+        | 4-functional group: 0 to 5
+        | 2-functoinal group: 6 to 17
```
lukasturcani, line 30: typo

lukasturcani (changes requested):

src/stk/molecular/topology_graphs/macrocycle/macrocycle.py
```diff
@@ -15,6 +15,9 @@ class Macrocycle(TopologyGraph):
     """
     Represents a macrocycle topology graph.
 
+    Building blocks with at least two functional groups are required
```
lukasturcani, line 18: must be exactly two

src/stk/molecular/topology_graphs/cage/metal_topologies/m4l6_tetrahedron.py
andrewtarzia: Lines 20 and 21 give the required information in this case as all vertices are the same.

lukasturcani: ah yeah, i guess we wanna remove this paragraph then?

andrewtarzia: mmm, either remove or I can see the benefit of it looking the same as the rest but with only one line - thoughts?

lukasturcani: Yes, I can also see the benefit in that, in which case I guess the paragraph above it would be edited. Certainly from the point of view of someone quickly scanning the documentation, I think that this option is probably better, but I don't have a strong opinion and will leave it to you to decide.

lukasturcani (approved):

PR #264, opened 2020-12-17
Cof connectivity

andrewtarzia:
Finish #205


src/stk/molecular/topology_graphs/cof/hexagonal.py
```diff
@@ -19,6 +19,15 @@ class Hexagonal(Cof):
     Building blocks with six and two functional groups are required
     for this topology graph.
 
+    When using a :class:`dict` for the `building_blocks` parameter,
+    as in :ref:`cage-topology-graph-examples`:
+    *Multi-Building Block Cage Construction*, a
```
lukasturcani, line 24: I don't think this should point to the cage examples.

lukasturcani (changes requested):

src/stk/molecular/topology_graphs/cof/hexagonal.py
andrewtarzia: I thought about this too - my logic here was that the example is effectively the same, regardless of material class - What is the alternative?

andrewtarzia: FIXED :D

lukasturcani (approved):

Issue #260, opened 2020-12-07
Mapping atoms from constructed molecule to building blocks


joshkamm:
Is there any available mechanism for mapping an atom in a constructed molecule to the corresponding atom in the building block that it was constructed from? If not, any pointers on how this could be done?

My initial use case is for linear polymers which may make things simpler but I could envision potentially going beyond that in the future.

If it would be helpful I can provide more details about what I'm trying to do.

P.S. I've been working on machine learning of good conformers of polymers and your library has been helpful. Thanks!

lukasturcani:
Hi & thank you! Yes this is possible. 

```python

constructed = stk.ConstructedMolecule(...)
for atom_info in constructed.get_atom_infos():
    atom = atom_info.get_atom()
    # If the atom was added during construction, these will return None.
    building_block = atom_info.get_building()
    building_block_id = atom_info.get_building_block_id()

```

https://stk.readthedocs.io/en/latest/stk.molecular.molecules.constructed_molecule.html#stk.molecular.molecules.constructed_molecule.ConstructedMolecule.get_atom_infos

https://stk.readthedocs.io/en/latest/stk.molecular.atoms.atom_info.html#stk.molecular.atoms.atom_info.AtomInfo

Key point is between `building_block` and `building_block_id`. Since each building block can be placed multiple times during the construction just returning the original building block might not be all the info you want. So building block id returned a unique id for every placement of a building block. For polymers, this means that you can identify each monomer, rather than just all the atoms in the constructed molecule which came from the original building block, which could be half of all the atoms in the polymer. (I feel like I always struggle to explain this, so let me know if its not clear).

There is also `BondInfo`, which gives you the same data but for bonds.

https://stk.readthedocs.io/en/latest/stk.molecular.molecules.constructed_molecule.html#stk.molecular.molecules.constructed_molecule.ConstructedMolecule.get_bond_infos

https://stk.readthedocs.io/en/latest/stk.molecular.bonds.bond_info.html#stk.molecular.bonds.bond_info.BondInfo

Might be good to include this example in the basic examples actually.

joshkamm:
Thanks for the informative response!

I think the building_block vs building_block_id makes sense partially because I already have to think about that distinction. I'm trying to use the atoms to map torsions between a polymer and its building blocks. An individual torsion could include atoms that span multiple occurrences of the same building block.

[...]

PR #267, opened 2020-12-24
Add AtomInfo.get_building_block_atom()

lukasturcani:
Should address https://github.com/lukasturcani/stk/issues/260

Main things:

* Add `AtomInfo.get_building_block_atom()`.
* Add `Molecule.get_canonical_atom_ids()`.

Adding `Molecule.get_canonical_atom_ids()` is a necessary evil, as far as I can tell. Here is why:

* `ConstructedMolecule.with_canonical_atom_ordering()` must work.
* When `ConstructedMolecule.with_canonical_atom_ordering()` is called, it also calls `Molecule.with_canonical_atom_ordering()` on its building blocks. This means the building blocks of a canonical constructed molecule are replaced with canonical building blocks.
* After `ConstructedMolecule.with_canonical_atom_ordering()` is called, `AtomInfo.get_building_block()` points to the new, canonical building block.
* Similarly `AtomInfo.get_building_block_atom()` needs to point to the atom in the new, canonical building block.
* The implementation of `ConstructedMolecule.with_canonical_atom_ordering()` which creates new `AtomInfo` instances for the new, canonical constructed molecule, needs to take the `Atom` from the old `AtomInfo` and figure out what the corresponding `Atom` in the new, canonical building block is. For this, the building block needs to provide a map, which maps the old id to the canonical id.

I will note that there is a slight inefficiency which I'm not too happy about. Canonical ordering is computed twice for each building block. Once when `with_canonical_atom_ordering()` is  called on it, and once when `get_canonical_atom_ids()` is called. This can be addressed in two ways as far as I can see:

1. Cache the ordering on first computation. Costs space.
2. Change the interface, such that when `with_canonical_atom_ordering` is called, it returns both the canonical clone, as well as the map from old to new id.

I prefer option two, because there is no performance penalty in time or space. Potentially this could be added as a separate method, so we would end up with 1 method for returning just clone, 1 for returning just map of ids, 1 for returning both together. Not sure if it would be worth keeping all three methods indefinitely or just slowly using deprecating 2 of the methods until we just end up with the method that returns both the clone and the map at the same time.


lukasturcani:
@andrewtarzia

andrewtarzia:
Seen - will do review.

src/stk/serialization/json/serializers/constructed_molecule/constructed_molecule.py
```diff
@@ -107,9 +107,17 @@ def get_keys(building_block):
...
+                return (
+                    None,
+                    None,
+                    None,
+                )
+
             return (
                 building_block_indices[atom_info.get_building_block()],
                 atom_info.get_building_block_id(),
+                atom_info.get_building_block_atom().get_id(),
```
andrewtarzia, line 120: I assume that this ordering does not matter compared to its ordering in the arguments list?

lukasturcani: Correct, and it means that backwards compatibility of the serialized representation is maintained.

andrewtarzia (approved):
Only question: is there a test that checks the ordering of a molecule constructed on the fly by the tests?

src/stk/serialization/json/serializers/constructed_molecule/constructed_molecule.py
andrewtarzia: Perfect. Good coding! ahha

lukasturcani:
I think so, if I understand your question the the following tests should cover it?
https://github.com/lukasturcani/stk/pull/267/files#diff-ffe064a35e027491836434ac1a055bf1ec9e262ca5524c682268fe7e4bc99709

andrewtarzia:
Ya I Think they do - just being cautious about the multi-steps being tested, but I think they should be fine!

lukasturcani:
Multi-steps?

andrewtarzia:
I am always concerned when we have an "old" case and a "new" case when the "new" case can undergo an arbitrary number of processes (i.e. reaction, canonicalization....). I think it is fine in this case, but in my head it is hard to picture all of this so I get nervous about whether its covered.

lukasturcani:
I see. Yeah I think that is a good instinct.

Issue #156, opened 2020-05-15
Add unaligning vertices


lukasturcani:
Most topologies at the moment expect functional groups with different placer ids. They can be made more tolerant by using a vertex which only places and doe not do any alignment, for cases where there is only 1 placer id in the entire molecule, across all functional groups.

Maybe warn the user when it is used?

lukasturcani:
@andrewtarzia - is gonna take the Linear class

andrewtarzia:
@lukasturcani the linear class already had an unaligning vertex, so just had to implement the replacement of linear vertices with an unaligning vertex if placer_ids is 1 -- this should be similar for many classes.

andrewtarzia:
@lukasturcani I would say we have covered this?

lukasturcani:
Agreed!

PR #275, opened 2021-01-03
Change cof vertex and add unaligned vertex for COFs

andrewtarzia:
Modify the alignment process for LinearCofVertex - align the normal of the plane of best fit through the entire BuildingBlock with the z-axis (therefore, hard-coded for 2D extended materials), then align the FG-FG vector along the edge-edge vector.

Handles issue #156 for Cof topologies.


lukasturcani:
Also would be good to add a test case where a building block has 3 functinal groups and 2 unique placer ids.

lukasturcani (approved):

andrewtarzia:
Bump - this is the next PR

Issue #277, opened 2021-01-04
Dump test structures to a MongoDB


lukasturcani:
At the moment, test structures get written to files. This is slightly inconvenient because the names of the files are so long, they get truncated so you do not even see which test generated it. Also through all of the structures is a pain, and requires some visualizer and lots of them are temperamental. 

The following things should be added:

* In addition to wirting the .mol files, also store the molecule in a `MoleculeMongoDb` and provide a description for it in a `ValueMongoDb`. 
* The name of the database, needs be unique to the pytest run, to prevent overwriting. See if pytest provides some kind of unique hash, or use a timestamp or a uuid.
* Make the CI dump the database binary file so that it can be downloaded and viewed

lukasturcani:
Solved by #322.

PR #322, opened 2021-04-29
Dump database

lukasturcani:
Related Issues: #277 

This PR adds a new test, `test_dump_molecule_structures()`, which 
is responsible for dumping the molecular structures to files and
placing them into a MongoDB database. Previously the file dumping
was done in the `test_get_atoms()` test, but I think having a 
separate test for this is better, because it keeps better to the single
responsibility principle. 

The PR also updates the CI to dump the MongoDB database after
the tests have been run, so that it can be downloaded and viewed
locally, for example with `stk-vis`. To help viewing with `stk-vis`
the file `tests/mongo-config.json` has been added. It can be 
loaded into `stk-vis` and configures it so that the developer does
not have to manually change the database name or the molecule
key name.

The PR also adds a new utility script `drop_test_database.py` which
just drops the databases which running the tests creates. This script
is meant to be run manually if a developer wants to remove the 
databases for some reason.

Finally, the PR also removes all uses of the MongoDB action with a 
MongoDB service container in the CI. This just seems more idiomatic
to me.


tests/drop_test_database.py
```diff
@@ -0,0 +1,24 @@
+"""
+A utility for dropping the databases created by running the tests.
+
+This script is not part of the test suite. It is meant to the run
```
andrewtarzia, line 4: be not the

andrewtarzia (commented):
Looks good. love this feature. One typo to fix, otherwise stunning.

Please clarify that local testing does not require MongoDB fully installed?

lukasturcani:
It does require it, but has for a while now.

Issue #330, opened 2021-05-19
Alignment issues in Macrocyles from two building blocks.


andrewtarzia:
When making a Macrocycle from only two building blocks, there is no handling of edge assignment when both edges are on the same position (calculated in Macrocyle.py on init based on the number of bbs in the chain).

MWE on my machine shows crossing bonds:
```python
import stk

bb1 = stk.BuildingBlock(
    'BrCCOCc1ccc(COCCBr)cc1',
    functional_groups=[stk.BromoFactory()],
)
bb2 = stk.BuildingBlock(
    'BrCOBr',
    functional_groups=[stk.BromoFactory()],
)
macrocycle = stk.ConstructedMolecule(
    topology_graph=stk.macrocycle.Macrocycle(
        building_blocks=(bb1, bb2),
        repeating_unit='AB',
        num_repeating_units=1,
    )
)
macrocycle.write('macrocycle.mol')
```

andrewtarzia:
An additional issue, but not a breaking issue.

Building blocks are not aligned such that their core atoms are away from the centre of the topology. This can be fixed with something like this for `class _CycleVertex(Vertex):`:

```
    def place_building_block(self, building_block, edges):
        assert (
            building_block.get_num_functional_groups() == 2
        ), (
            f'{building_block} needs to have exactly 2 functional '
            'groups but has '
            f'{building_block.get_num_functional_groups()}.'
        )
        building_block = building_block.with_centroid(
            position=self._position,
            atom_ids=building_block.get_placer_ids(),
        )
        fg0, fg1 = building_block.get_functional_groups()
        fg0_position = building_block.get_centroid(
            atom_ids=fg0.get_placer_ids(),
        )
        fg1_position = building_block.get_centroid(
            atom_ids=fg1.get_placer_ids(),
        )
        building_block = building_block.with_rotation_between_vectors(
            start=fg1_position - fg0_position,
            target=[-1 if self._flip else 1, 0, 0],
            origin=self._position,
        )
        building_block = building_block.with_rotation_about_axis(
            angle=self._angle-(np.pi/2),
            axis=np.array([0, 0, 1]),
            origin=self._position,
        )

        # Orient building blocks away from centre of cycle.
        core_centroid = building_block.get_centroid(
            atom_ids=building_block.get_core_atom_ids(),
        )
        edge_centroid = (
            sum(edge.get_position() for edge in edges) / len(edges)
        )

        building_block = building_block.with_rotation_between_vectors(
[...]

PR #332, opened 2021-05-19
Add ValueError for Macrocycle from 2 BBs

andrewtarzia:
Related Issues: #330

This PR introduces an error to let the user know (in a hard way, because the output is nonsense) that alignment is not supported for Macrocycles formed from only 2 building blocks and 1 repeat unit. This is an edge use-case that I think can be fixed, with large rewrites of the Macrocycle class. A 2 building block Macrocycle also makes less sense, then using a different topology for that molecule.


lukasturcani:
Whats the issue when two building blocks are used?

lukasturcani (approved):
Ty! LGTM.

Issue #288, opened 2021-03-25
Linear confused by only one building block and repeat unit


joshkamm:
I found that I was getting an error similar to this:

```python
~/miniconda3/lib/python3.7/site-packages/stk/molecular/topology_graphs/polymer/linear/linear.py in __init__(self, building_blocks, repeating_unit, num_repeating_units, orientations, random_seed, reaction_factory, num_processes)
    251         self._num_repeating_units = num_repeating_units
    252 
--> 253         head, *body, tail = orientations
    254         choices = [True, False]
    255         vertices = [

ValueError: not enough values to unpack (expected at least 2, got 1)
```

After a bit of debugging I realized that I had multiple layers of construction and the error seemed to occur when one of the layers was functioning as a constant layer (constructing a linear polymer with only one monomer). I verified the hypothesis with this adaptation of an example from the docs which produced the error above:

```python
import stk

# React the amine functional groups during construction.
bb1 = stk.BuildingBlock('NCCN', [stk.PrimaryAminoFactory()])
# React the aldehyde functional groups during construction.
bb2 = stk.BuildingBlock('O=CCCC=O', [stk.AldehydeFactory()])
# Build a polymer.
polymer = stk.ConstructedMolecule(
    topology_graph=stk.polymer.Linear(
        building_blocks=(bb1, bb2),
        repeating_unit='A',
        num_repeating_units=1,
    ),
)
```

There is no error if `repeating_unit == 'AB'` or if `num_repeating_units == 2`. It seems like `Linear` isn't prepared for the edge case of `num_repeating_units == 1` and `repeating_unit` only having one building block. In basic usage it makes sense that you wouldn't bother constructing a molecule that's the same as one of your building blocks but it'd be convenient to not have to think about that as a special case when it occurs in multi-layered construction.

lukasturcani:
Thanks for reporting this. Yeah, there is no real need for the edge case to exist, so  I agree it should be removed.

lukasturcani:
This is now available in the new stk release, `v2021.5.26.0`,  available on pip, so existing stk installations can be updated with `pip install stk --upgrade`.

PR #338, opened 2021-05-21
Allow linear polymers from 1 building block

lukasturcani:
Related Issues: #288

@joshkamm feel free to have a look through / review this PR too
if you find the time. If you wanna comment (or not comment) after it's
been merged, that's OK too.

Users can now create a `Linear` polymer which has only 1 building
block. The implementation relies on adding a new vertex,
`_UnaligningVertex` which just places the building block. The idea
being that when there is only 1 building block, there is no need to
do any aligning.

In order to implement this change I factored out most of the old
initializer into a separate function, `_get_vertices_and_edges`. The
new initializer tries to unpack the list of orientations into a head,
body and  tail. If this succeeds, `_get_vertices_and_edges` will get
called. If it fails, it means that there is only 1 building block so
an `_UnaligningVertex` gets made instead.

The `_UnaligningVertex` has it's flipping always set to `False`. This
will  happen even if the probability of flipping is set to 1. This is
weird, but it simplifies the implementation and I do not think it
matters in practice. In any case, I have documented this behavior in
the docstring. If the flipping was allowed, it would mean that
`_UnaligingVertex` would actually have to start doing orientations, or
flips along some  axis, which I think is completely unnecessary.

The only way a user would notice this weirdness is if they print the
string representation of the topology graph. So one option could be
to let the representation show that the vertex does flipping, but
then not actually do it. I think this would also be weird, because if
the user checked if the position matrix had changed, it would not have,
and that's confusing too! So I think out of the options of doing
something completely unnecessary, versus having a string
representation that might not match the initialization parameters,
I think the latter is way better, because I really doubt anyone will
notice.

There is also the option of having representation that does not
show orientations for the case when the polymer is only made
from one building block. However, even in this case the user
might expect that by setting the orientations parameter to 0 or 1
different positions matrices will result. I could just prevent the
use of the orientation parameter for this case, but again I think
that's just too complicated overall.

There is also a minor change of moving the `polymer_linear_opt`
into just the `polymer_linear` test case. I don't see why a separate
fixture was made for `polymer_linear_opt`, I think it's just another
test case of `polymer_linear` that has no reason to be separated
from its friends.


src/stk/molecular/topology_graphs/polymer/linear/linear.py
```diff
@@ -245,6 +261,12 @@ def __init__(
             length of `orientations` must be equal to
             ``len(repeating_unit)*num_repeating_units``.
 
+            If there is only one building block in the constructed
+            polymer i.e. the `repeating_unit` has a length of 1 and
```
andrewtarzia, line 265: this is a little unclear, should be ."i.e. there is only one building block, the `repeating_unit` ....."

andrewtarzia (approved):
Super minor change. Solves the issue.

PR #339, opened 2021-05-21
Plug PeriodicInfo state leak

lukasturcani:
Related Issues: #307 

When `PeriodicInfo` is being initialized it is not making a copy of
the arrays it is being given. This means thats if those arrays get
changed on the outside, the internal state of `PeriodicInfo` also gets
corrupted. Similarly, when the `get` methods are called, the internal
state is returned directly, rather than a copy of the arrays. If
someone calls `get` and then changes the value in the returned array,
they will also change the internal state of `PeriodicInfo`.

This commit just makes sure to make a copy of the arrays which are
received as input and returned as output.

### DIFF
--- a/src/stk/molecular/periodic_info.py
+++ b/src/stk/molecular/periodic_info.py
@@ -43,10 +43,15 @@ class PeriodicInfo:
         .. [1] https://pymatgen.org/_modules/pymatgen/core/lattice.html
 
         """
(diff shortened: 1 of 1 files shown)


andrewtarzia (approved):
Did not realise np.array created a clone, rather than a state change, but this is a great find! Thank you!

Issue #333, opened 2021-05-20
Create Docker Container for Running Tests in CI


lukasturcani:
CI still seems to not lead to reproducible steps, so it looks like a container has  to be used. Worth exploring if GitHub Packages will work for this.

PR #342, opened 2021-05-23
Add test container

lukasturcani:
Related Issues: #333

The CI tests now run in a container,
`ghcr.io/lukasturcani/stk-test-environment`. The reason for this is
primarily that we have a flaky test at the moment and ensuring a
consistent environment isolates the possible causes. A secondary cause
is that by having all the pre-requisites for tests installed in a
conatiner already CI speed improves. This speed improvement is pretty
small because the CI was pretty fast already.

The container which is used for testing is created by the Dockerfile
added to `tests/docker/Dockerfile`, and can be created by running
`docker build -t stk-test-environment .` inside the `tests/docker`
folder. The container image has been added to GitHub packages for easy
access.

When tests are run in a container, the MongoDB database can no longer
be accessed on `localhost`. This means that the current tests, both
pytest and doctest, which assume that a database exists on `localhost`
needed to be updated. For the pytest tests, a new option was added so
that the database the tests want to connect to can be configured
through a command line argument
`pytest --mongodb_uri=--mongodb_uri='mongodb://MongoDB:27017/'`. For
doctest tests an environment variable is used to change the URI the
tests want to connect to
`MOCK_MODULES='FALSE' MONGODB_URI='mongodb://MongoDB:27017/'
make -C docs doctest`.

In order to allow pytest tests to easily switch the database to which
they connect, I added a new fixture which reads the new command line
option and initializes a `MongoClient` instance using the URI from the
command line. The tests, and test cases, then rely on this fixture
rather than creating the `MongoClient` instance themselves. One funny
thing is that I introduced a bunch of `CaseDataData` classes, which
have a weird name, but I could not think of a better one. The existing
`CaseData` instances hold a database object, which means that they
need to be constructed using a `MongoClient` instance, because this
is a required parameter to the MongoDB database initializers. However,
the new `MongoClient`instance is a fixture, so I need to be inside
a test or a fixture to have access to it. This means if I am
parametrizing a `CaseData` fixture, I don't have access the
`MongoClient` instance I need to create `CaseData` instances. To get
around this, I've created the `CaseDataData` classes. Instead of
holding the database, they hold a `get_database` function which, when
given the `MongoClient`, creates the database. So now I can parametrize
my fixtures using `CaseDataData`, the fixture itself can depend on the
new `MongoClient` fixture, and the body of parameterized fixture can
create a `CaseData` by combining the `get_database` with the
`MongoClient`. This can be seen in
`tests/databases/molecule/fixtures/molecule_mongo_db.py`.

The implementation of doctest tests was less neat. I'm using an
environment variable to input the URI. I don't know of a better way 
to send options to the doctest tests, but to be honest I did not actually 
look very hard. The test setup replaces the `pymongo.MongoClient`
initializer with a custom function which initializes the `MongoClient`
with the URI provided in the environment variable. The reason for this
is that I want the example to show something like
`pymongo.MongoClient()` but still connect to a custom MongoDB endpoint
rather than `localhost`. By replacing the `pymongo.MongoClient`
initializer with my own in the test setup, I can make
`pymongo.MongoClient()` connect to the URI given in the environment
variable. The teardown for the tests was updated so that the original
`pymongo.MongoClient` initializer is restored after the test is run.

I should note that this commit does not actually fix the flaky test,
but it is a step in the right direction. So I'm going to merge this
and continue looking into it.


.github/workflows/remake_rst_files.yml
```diff
@@ -6,18 +6,21 @@ on:
 
 jobs:
     remake-rst-files:
-        runs-on: ubuntu-latest
+        runs-on: ubuntu-20.04
```
andrewtarzia, line 9: Seems unnecessary to set the ubuntu version away from latest? Should then run tests on multiple versions?

.github/workflows/tests.yml
```diff
@@ -4,135 +4,105 @@ on: [push, pull_request]
 
 jobs:
     lint:
-        runs-on: ubuntu-latest
+        runs-on: ubuntu-20.04
```
andrewtarzia, line 7: as above.

.github/workflows/tests.yml
```diff
@@ -4,135 +4,105 @@ on: [push, pull_request]
...
+              run: conda run -n stk
                 flake8 .
                     --count
                     --show-source
                     --max-line-length=71
                     --statistics
 
     pytest-stable-linux:
-        runs-on: ubuntu-latest
+        runs-on: ubuntu-20.04
```
andrewtarzia, line 24: as above.

.github/workflows/tests.yml
```diff
@@ -4,135 +4,105 @@ on: [push, pull_request]
...
             - name: Upload the database dump
+              if: ${{ always() }}
               uses: actions/upload-artifact@v2
               with:
                 name: database-dump
                 path: database-dump
 
     doctests:
-        runs-on: ubuntu-latest
+        runs-on: ubuntu-20.04
```
andrewtarzia, line 59: as above.

.github/workflows/tests.yml
```diff
@@ -147,38 +117,31 @@ jobs:
                 path: rotatable_bonds_progress.png
 
     intermediate-ea:
-        runs-on: ubuntu-latest
+        runs-on: ubuntu-20.04
```
andrewtarzia, line 120: as above.

.github/workflows/tests.yml
```diff
@@ -4,135 +4,105 @@ on: [push, pull_request]
...
-              run:
-                MOCK_MODULES='FALSE' make -C docs doctest
+              run: conda run -n stk
+                MOCK_MODULES='FALSE'
+                MONGODB_URI='mongodb://MongoDB:27017/'
+                make -C docs doctest
 
     basic-ea:
-        runs-on: ubuntu-latest
+        runs-on: ubuntu-20.04
```
andrewtarzia, line 82: as above.

src/stk/databases/mongo_db/constructed_molecule.py
```diff
@@ -48,6 +48,18 @@ class ConstructedMoleculeMongoDb(ConstructedMoleculeDatabase):
             )
         )
 
+        # Change the database MongoClient will connect to.
```
andrewtarzia, line 51: Is this snippet in all examples now? Just seeing it pop up a lot?

tests/databases/constructed_molecule/fixtures/constructed_molecule_mongo_db.py
```diff
@@ -2,18 +2,48 @@
 import stk
 import rdkit.Chem.AllChem as rdkit
 import pymongo
+from dataclasses import dataclass
+from typing import Callable
```
andrewtarzia, line 6: Safe. I like it ;)

andrewtarzia (commented):
My main thought is that this PR is a good time to have some documentation on writing tests and looking at this database, for example, as it is getting quite complicated!

.github/workflows/remake_rst_files.yml
lukasturcani: I'm pinning the version so that we can be sure what our environment is. I don't want the same command to suddenly start creating new images.

src/stk/databases/mongo_db/constructed_molecule.py
lukasturcani: In all the examples that use `MongoClient`

lukasturcani:
Yeah I agree 100% with your comment about documentation of the CI. Will do it in a separate PR but defo would like to get it done sooner rather than later.

.github/workflows/remake_rst_files.yml
lukasturcani: To be fair, I  think the image might end up being slightly different anyway, because I do not pin the version on thing I install via `apt-get`.

andrewtarzia: Maybe, once the issue is fixed, we remove this?

andrewtarzia: Maybe, once the issue is fixed, we remove this?

andrewtarzia:
I approve though!

.github/workflows/remake_rst_files.yml
lukasturcani: Nah I mean I want that guarantee irrespective of the issue. I guess I'm not clear on the benefit of using latest.

Issue #344, opened 2021-05-28
Accessing private vertex classes 


andrewtarzia:
All implemented vertex classes are "private" with names like "_LinearCageVertex". This makes them difficult to access by users who may want to implement a new topology graph in their own code without reimplementing the vertex classes (which makes no sense).

Suggestion is to make them as accessible as other classes; i.e. 

```python
import stk

vertex = stk.cage.LinearCageVertex(...)
```

lukasturcani:
I guess I can also just rename it to `stk.cage.LinearVertex` when I do this.

PR #346, opened 2021-06-04
Vertex name changes

andrewtarzia:
Related Issues: #344

This PR makes all vertex classes (except for Cage and Cof base classes) publicly accessible by simply changing class names. Also removes designation of Cage and Cof types within name, as that is designated by the access point. (sorry if my terminology here is wrong). An example of accessing the vertices was added.

The following example shows that this works.

```python
import stk

vertex = stk.Vertex(0, [0, 0, 0])
print(vertex)
try:
    vertex = stk.LinearVertex(1, [0, 0, 0])
    print(vertex, 'should not show')
except AttributeError:
    print('success on fail')
vertex = stk.cage.LinearVertex(1, [0, 0, 0])
print(vertex)
vertex = stk.cage.vertices.LinearVertex(1, [0, 0, 0])
print(vertex)
vertex = stk.cage.OnePlusOneVertex(1, [0, 0, 0], 0)
print(vertex)
vertex = stk.polymer.linear.TailVertex(0, [0, 0, 0], False)
print(vertex)
vertex = stk.cof.NonLinearVertex(0, [0, 0, 0], False)
print(vertex)
vertex = stk.host_guest.HostVertex(0, [0, 0, 0])
print(vertex)
vertex = stk.macrocycle.CycleVertex(0, [0, 0, 0], False, 0)
print(vertex)
vertex = stk.metal_complex.MonoDentateLigandVertex(0, [0, 0, 0])
print(vertex)
vertex = stk.rotaxane.AxleVertex(0, [0, 0, 0])
print(vertex)
try:
    vertex = stk.AxleVertex(1, [0, 0, 0])
    print(vertex, 'should not show')
except AttributeError:
    print('success on fail')
try:
    vertex = stk.HostVertex(1, [0, 0, 0])
    print(vertex, 'should not show')
except AttributeError:
    print('success on fail')
```


docs/source/basic_examples.rst
```diff
@@ -1022,6 +1022,23 @@ useful key for metal-containing molecules. You can use the
...
+========================
+
+The vertex classes that make up toplogy graphs in :mod:`stk` can be
+accessed if required.
+
+.. testcode:: accessing-vertex-classes
+
+    import stk
+
+    base_vertex = stk.Vertex(0, [0, 0, 0])
```
lukasturcani, line 1035: I think the example should demonstrate to the user how to solve a specific problem. So in this case, maybe like make a new cage topology graph? 

I thiink giving an example of making a `stk.Vertex` instance is bad, because you actually never want to make an instance of it -> it has unimplemetned methods. You only use it for the purposes inheritence, when defining a new Vertex class.

docs/source/basic_examples.rst
```diff
@@ -1022,6 +1022,23 @@ useful key for metal-containing molecules. You can use the
     stk.MoleculeMongoDb = _old_molecule_init
     pymongo.MongoClient = _mongo_client
 
+Accessing Vertex classes
```
lukasturcani, line 1025: This ain't title case.

docs/source/basic_examples.rst
```diff
@@ -1022,6 +1022,39 @@ useful key for metal-containing molecules. You can use the
     stk.MoleculeMongoDb = _old_molecule_init
     pymongo.MongoClient = _mongo_client
 
+Accessing Vertex Classes
```
lukasturcani, line 1025: I would probably rename this something like "Creating New Topology Graphs with Existing Vertices" or sometrhing better if you like

docs/source/basic_examples.rst
```diff
@@ -1022,6 +1022,39 @@ useful key for metal-containing molecules. You can use the
...
+========================
+
+The vertex classes that make up toplogy graphs in :mod:`stk` can be
```
lukasturcani, line 1028:
```
:mod:`.stk`
```

docs/source/basic_examples.rst
```diff
@@ -1022,6 +1022,39 @@ useful key for metal-containing molecules. You can use the
...
+accessed if required. For example, this allows the user to define a
```
lukasturcani, line 1029: can be accessed  to speed up the implemention of new topology graphs

docs/source/basic_examples.rst
```diff
@@ -1022,6 +1022,38 @@ useful key for metal-containing molecules. You can use the
...
+
+    class NewCageTopology(stk.cage.Cage):
+
+        _vertex_prototypes = (
+            stk.cage.LinearVertex(0, [0, 0, 0]),
+            stk.cage.LinearVertex(1, [0, 1, 0]),
+        )
+
+        # Define Edges below.
+        _edge_prototypes = ()
```
lukasturcani, line 1043: I think these examples should contain *some* edges, because that is an important part of creating new topology graphs /  cages. I think omitting them can be a little misleading.

tests/molecular/topology_graphs/vertex/cases/cage/conftest.py
```diff
@@ -82,9 +82,9 @@ def cls(request):
 
 @pytest.fixture(
     params=(
-        vertices._LinearCageVertex.init_at_center,
-        vertices._NonLinearCageVertex.init_at_center,
-        vertices._UnaligningVertex.init_at_center,
+        vertices.LinearVertex.init_at_center,
```
lukasturcani, line 85: now that the vertices are public, I think they should be accessed in tests by the public API, not by this "unofficial" API. so `stk.cage.LinearVertex` not `vertices.LinearVertex`.

lukasturcani (changes requested):

lukasturcani:
I also have some concerns about exposing *all* of the vertex classes. I'm not against it, but I guess I would like to know what the arguments for and against this are? For example, what is the cost / benefit of exposing `stk.host_guest.HostVertex`?

docs/source/basic_examples.rst
andrewtarzia: Done.

andrewtarzia:
I see your concern and my main argument is how do you decide which to make public and which not to? There is no technical reason I see for that other than the desire to minimise access until needed.

For example, why make the cage vertex accessible, but not the host vertex? A user may be just inclined to use that in their own code as an `stk.cage.LinearVertex`

tests/molecular/topology_graphs/vertex/cases/cage/conftest.py
andrewtarzia: I agree and have replaced all `vertices = stk.cage.vertices` (for example) with direct calls to `stk.cage.VERTEXNAME`.

lukasturcani:
I guess with cages the extensibility argument is a bit easier. The amount of cage topologies we have already, means that we have a pretty good understanding of how this vertex can be used to easily build new topology graphs. With something like host guest, we have experience in using it in 1 topology graph. so the reusability of it a lot less well understood. I guess I'm not convinced that just because a user wants to use that vertex, that I want to allow them to, because doing so would remove my flexibility to modify it in the future.

Having said that, I think we can expose them because removing them can be easy for the users to fix (just copy paste the code into their own scripts). So I think it is a worthwhile experiment.

docs/source/basic_examples.rst
```diff
@@ -1022,6 +1022,43 @@ useful key for metal-containing molecules. You can use the
...
+accessed to speed up the implemention of new topology graphs.
+
+.. testcode:: creating-new-topology-graphs-with-existing-vertices
+
+    import stk
+
+    class NewCageTopology(stk.cage.Cage):
+
+        _vertex_prototypes = (
+            stk.cage.LinearVertex(0, [0, 0, 0]),
```
lukasturcani, line 1038: Also, I think there is a type error here. the position should be a `tuple` of `float` so `(0., 0., 0.)`

https://stk.readthedocs.io/en/stable/stk.molecular.topology_graphs.topology_graph.vertex.html#stk.molecular.topology_graphs.topology_graph.vertex.Vertex

I realize that this type error exists in the actual code, but I don't want you to change that. This PR just needs to make sure that users get a fully correct example.

lukasturcani (changes requested):

docs/source/basic_examples.rst
andrewtarzia: Done.

lukasturcani (approved):

PR #352, opened 2021-06-23
First attempt at multiple guests.

andrewtarzia:
Related Issues: #351 
Requested Reviewers: @lukasturcani 

This PR is the beginnings of a new feature for the host_guest topology graph that allows for any number of guests to be placed inside a host. Currently, this is done by modifying the Complex API to take an iterable of `guests`, which are `.Guest(TopologyGraph)` classes. Unsure if that is the best approach, but it provides the needed methods and avoids i) a new type of class for the Guests ii) or expanding the `BuildingBlock` class or iii) an odd usage of a vertex as input to a TopologyGraph through a `.GuestVertex` class (my initial thought but nowhere else in stk is this done). 

Also tempted to have a `.Complex` and `.NGuestComplex` class so that the original method can be maintained for backwards compatibility.

Submitting this to get feedback, the placement of code in stk, tests, examples and docstrings are not complete.

* `src/stk/molecular/topology_graph/host_guest/complex.py:Complex.init()`: This method is simplified greatly and simply handles the transform of host and guests arguments into `building_block_vertices`
* 
* `src/stk/molecular/topology_graph/host_guest/complex.py`: Add new class (temporarily), `.Guest(TopologyGraph)` whose `init` is effectively equivalent to the old `Complex.init()` and handles the definition of the properties required to define a `GuestVertex`.

See the code below for building a structure with 1 or 2 guests:

```python

import stk

host = stk.ConstructedMolecule(
    topology_graph=stk.cage.FourPlusSix(
        building_blocks=(
            stk.BuildingBlock('BrCCBr', [stk.BromoFactory()]),
            stk.BuildingBlock(
                smiles='BrCC(Br)CBr',
                functional_groups=[stk.BromoFactory()],
            ),
        ),
    ),
)
host = stk.BuildingBlock.init_from_molecule(host)
guest1_bb = stk.BuildingBlock('c1ccccc1')
guest1 = stk.host_guest.Guest(
    guest=guest1_bb,
    guest_start=guest1_bb.get_direction(),
    guest_target=(1., 0., 0.),
    displacement=(0, 0, 0),
)
guest2_bb = stk.BuildingBlock('C1CCCC1')
guest2 = stk.host_guest.Guest(
    guest=guest2_bb,
    guest_start=guest2_bb.get_direction(),
    guest_target=(0., 1., 0.),
    displacement=(0, 0, 0),
)

print(host, guest1, guest2)

complex = stk.ConstructedMolecule(
    topology_graph=stk.host_guest.Complex(
[...]

PR #356, opened 2021-06-30
Implement Spinner optimizer with docs and tests.

andrewtarzia:

Simply add another optimizer called Spinner, which is made to optimize/generate conformers of host-guest pairs. This is most crucial for systems with multiple guests that may have overlapping structures.

I followed the same approach as the mchammer optimizers (effectively the same interface and code).

Adds a new dependency. It is small, but worth making the decision about its inclusion.

minimum working example (tests also included) - compare the two output structures.:

```python
import stk

host = stk.ConstructedMolecule(
    topology_graph=stk.cage.FourPlusSix(
        building_blocks=(
            stk.BuildingBlock('BrCCBr', [stk.BromoFactory()]),
            stk.BuildingBlock(
                smiles='BrCC(Br)CBr',
                functional_groups=[stk.BromoFactory()],
            ),
        ),
    ),
)
host = stk.BuildingBlock.init_from_molecule(host)
guest1_bb = stk.BuildingBlock('c1ccccc1')
guest1 = stk.host_guest.Guest(
    building_block=guest1_bb,
    start_vector=guest1_bb.get_direction(),
    end_vector=(1., 0., 0.),
    displacement=(0, 0, 0),
)
guest2_bb = stk.BuildingBlock('C1CCCC1')
guest2 = stk.host_guest.Guest(
    building_block=guest2_bb,
    start_vector=guest2_bb.get_direction(),
    end_vector=(0., 1., 0.),
    displacement=(0, 0, 0),
)

complex.write('complex_2.mol')
complex = stk.ConstructedMolecule(
    topology_graph=stk.host_guest.Complex(
        host=host,
        guests=(guest1, guest2),
    ),
)
complex.write('complex_3.mol')

complex = stk.ConstructedMolecule(
    topology_graph=stk.host_guest.Complex(
        host=host,
        guests=(guest1, guest2),
        optimizer=stk.Spinner(),
    ),
)
complex.write('complex_3_opt.mol')

```

### DIFF
--- a/src/stk/molecular/topology_graphs/host_guest/complex.py
+++ b/src/stk/molecular/topology_graphs/host_guest/complex.py
@@ -174,7 +174,7 @@ class Complex(TopologyGraph):
 
         complex = stk.ConstructedMolecule(
             topology_graph=stk.host_guest.Complex(
-                host=stk.BuildingBlock.init_from_molecule(host),
+                host=stk.BuildingBlock.init_from_molecule(cage),
                 guests=guest,
                 optimizer=stk.Spinner(),
             ),
--- a/src/stk/molecular/topology_graphs/host_guest/complex.py
+++ b/src/stk/molecular/topology_graphs/host_guest/complex.py
@@ -151,10 +151,34 @@ class Complex(TopologyGraph):
 
     *Suggested Optimization*
 
-    For :class:`.Complex` topologies, there is no need to use an
-    optimizer, so stick with :class:`.NullOptimizer`. However, it is
-    recommended that all building blocks be optimized prior to
-    construction.
+    For :class:`.Complex` topologies, it is recommended to use the
+    :class:`.Spinner` optimizer. It is also recommended that the
+    building blocks are already optimized prior to construction.
+
+    .. testcode:: suggested-optimization
+
+        import stk
+
+        bb1 = stk.BuildingBlock('NCCN', [stk.PrimaryAminoFactory()])
+        bb2 = stk.BuildingBlock(
+            smiles='O=CC(C=O)C=O',
+            functional_groups=[stk.AldehydeFactory()],
+        )
+        guest = stk.host_guest.Guest(stk.BuildingBlock('c1ccccc1'))
+        cage = stk.ConstructedMolecule(
+            topology_graph=stk.cage.FourPlusSix(
+                building_blocks=(bb1, bb2),
+                optimizer=stk.MCHammer(),
+            ),
+        )
+
+        complex = stk.ConstructedMolecule(
+            topology_graph=stk.host_guest.Complex(
+                host=stk.BuildingBlock.init_from_molecule(host),
+                guests=guest,
+                optimizer=stk.Spinner(),
+            ),
+        )
 
     *Changing the Position of the Guest*
 
--- a/src/stk/molecular/topology_graphs/host_guest/complex.py
+++ b/src/stk/molecular/topology_graphs/host_guest/complex.py
@@ -20,10 +20,10 @@ class Guest:
     def __init__(
         self,
         building_block: BuildingBlock,
-        start_vector: Tuple[float] = (1., 0., 0.),
-        end_vector: Tuple[float] = (1., 0., 0.),
-        displacement: Tuple[float] = (1., 0., 0.),
-    ):
+        start_vector: Tuple[float, float, float] = (1., 0., 0.),
+        end_vector: Tuple[float, float, float] = (1., 0., 0.),
+        displacement: Tuple[float, float, float] = (1., 0., 0.),
+    ) -> None:
         """
         Initialize a :class:`.Guest` instance.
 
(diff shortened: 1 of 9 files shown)


src/stk/molecular/topology_graphs/host_guest/complex.py
```diff
@@ -4,8 +4,17 @@
 
 """
 
+from typing import Dict, Optional, Tuple, Iterable
```
lukasturcani, line 7: Use 
```python
from __future__ import annotations
```

and then you can use `tuple, dict` for type hints instead of importing `typing.Tuple` and `typing.Dict`

src/stk/molecular/topology_graphs/host_guest/complex.py
```diff
@@ -118,152 +114,174 @@ class Complex(TopologyGraph):
...
+            complex = stk.ConstructedMolecule(
+                topology_graph=stk.host_guest.Complex(
+                    host=stk.BuildingBlock.init_from_molecule(host),
+                    guests=stk.host_guest.Guest(
+                        building_block=stk.BuildingBlock('[Br][Br]'),
+                    ),
                 ),
-            ),
-        )
+            )
```
lukasturcani, line 149: Maybe add an example here of using multiple guests?

src/stk/molecular/topology_graphs/host_guest/complex.py
```diff
@@ -374,15 +396,21 @@ def _get_vertices_from_guest(
 
         return building_block_vertices
 
-    def clone(self):
+    def clone(self) -> 'Complex':
```
lukasturcani, line 399: Can stop being a string once the annotations are imported from `__future__`.

src/stk/molecular/topology_graphs/topology_graph/optimizers/spinner.py
```diff
@@ -0,0 +1,146 @@
...
+                ),
+            )
+
+        Optimisation with :mod:`stk` simply collects the final position
+        matrix. The optimisation's trajectory can be output using the
+        :mod:`SpinDry` implementation if required by the user [1]_.
+        This code is entirely nonphysical and is, therefore, completely
+        general to any chemistry.
+
+    References
```
lukasturcani, line 64: Change to Google style heading

src/stk/molecular/topology_graphs/topology_graph/optimizers/spinner.py
```diff
@@ -0,0 +1,146 @@
...
+
+            rotation_step_size: The relative size of the rotation to
+                take during step.
+
+            num_conformers: Number of conformers to extract.
+
+            max_attempts: Maximum number of MC moves to try to generate
+                conformers.
+
+            nonbond_epsilon: Value of epsilon used in the nonbond
```
lukasturcani, line 96: Is it the non-bond potential or non-bonded potential? (or both?)

src/stk/molecular/topology_graphs/topology_graph/optimizers/spinner.py
```diff
@@ -0,0 +1,146 @@
...
+            num_conformers=num_conformers,
+            max_attempts=max_attempts,
+            nonbond_epsilon=nonbond_epsilon,
+            nonbond_sigma=nonbond_sigma,
+            beta=beta,
+            random_seed=random_seed,
+        )
+
+    def optimize(self, state: ConstructionState) -> ConstructionState:
+        # Define SpinDry supramolecule to optimize.
```
lukasturcani, line 122: Self-explanatory comment can be removed I think.

src/stk/molecular/topology_graphs/topology_graph/optimizers/spinner.py
```diff
@@ -0,0 +1,146 @@
...
+                    atom_ids=(
+                        bond.get_atom1().get_id(),
+                        bond.get_atom2().get_id(),
+                    )
+                ) for i, bond in enumerate(state.get_bonds())
+            ),
+            position_matrix=state.get_position_matrix(),
+        )
+
+        # Run optimization.
```
lukasturcani, line 142: This one also.

lukasturcani (changes requested):
This is amazing work, thank you so much @andrewtarzia. Just some very minor nits.

lukasturcani:
There are two warnings in the docs compilation, which I think were introduced by #352. Do you mind fixing them in this PR?

andrewtarzia:
> There are two warnings in the docs compilation, which I think were introduced by #352. Do you mind fixing them in this PR?

Could you send the info? - very happy to make the changes.

lukasturcani:
At the bottom of the github page you will have a "All checks have passed" row, which has a "Show all checks button". If you lick on that and scroll through the different checks, there should be one for readthedocs. If you click on it, it will take you to the docs. If you lick the "v:356"  in the bottom left corner and then on "builds". You can go through the builds until you find "356". there you will see the different steps. click on the last one and scroll of the way down. You will see a warning (there's only one tbh, not two, I think you must have fixed the previous ones already). You have to scroll up a bit to find the actual warning. 

The warning itself is because you didn't run remake-modules. WHich is fine cos thats autodamted now. 

So you dont actually have to do anything. But I figured I would post this for future reference.

andrewtarzia:
Ahhh amazing- I always forget about that. Will do other changes Wed.

src/stk/molecular/topology_graphs/topology_graph/optimizers/spinner.py
andrewtarzia: I think nonbonded is better.

lukasturcani:
@andrewtarzia actually you probably should run remake-modules, becaues without it you cannot see the compiled docs for Spinner, so its hard to make sure everything is OK.

src/stk/molecular/topology_graphs/topology_graph/optimizers/spinner.py
lukasturcani: How do you feel about chaning the argument name as well in that case?

lukasturcani (approved):

src/stk/molecular/topology_graphs/topology_graph/optimizers/spinner.py
andrewtarzia: Poorly because that is the argument name in SpinDry and I would want consistency between the two.

andrewtarzia: I think the argument/variable name is clear as to what it applies to.

lukasturcani: I don't think the consistency between the two is important. When stk wraps another libraries interface, it should not aim for consistency with it, because that would make stks own interface inconsistent. For example, stk cannot be both consistent with numpy, rdkit and SpindDry interfraces. it can only be consistent with its own.

I think it clear also, but if one form is better than the other, then I would prefer the better one, even if the first one is good.

lukasturcani: also stk has not control over the interface of its dependencies. This means spin dry might change its interface, and suddenly stk has a worse argument + its not even consistent any more.

andrewtarzia: I agree with your statement, however, consistency with spindry is trivial compared to rdkit.

Additionally, `nonbond_epsilon` is used in the MCHammer implementation. So internal consistency is maintained.

lukasturcani: Yep, if stk already uses `nonbond_x` then I agree this should follow the same pattern.

PR #357, opened 2021-06-30
Add setup for type hints

lukasturcani:

This commit allows `stk` to use type hints. The purpose of type hints
is to give us the opportunity to use static type checking. In order
to use type hints, the types must be removed from docstrings, so that
they are not specified in two places, which would be a maintenance
burden. However, the types should still be present in the compiled
documentation. By using `sphinx_autodoc_typehints`, the types in
type hints are still present in the compiled docs. However, because
the NumPy style docstring require the return type of a function to
be present, a switch is going to be made to Google style docstrings,
which do not have this requirement.

This commit updates the docstrings of the `Atom` class as an initial
example of the changes going forward. The type hints will be
introduced gradually. Every commit which touches a a function must
introduce type hints to it and switch the docstrings to Google style.
Use of `mypy` in CI will not be introduced in this commit, but in a
later one, once more of code base has type annotations.

In order to implement type hints for the `Atom` class a couple of
things were changed. First, the default implementation was moved
from the `Atom` class itself, into `AtomImpl`. This is because the
`init_subclass` method was changing the `__init__` method of the
subclass, which was causing a type error. The code was safe, but I
think overall more confusing than just simply using inheritance to
override the init method. I also explicitly overrode the clone method
for all the elements, in order to support covariant return types.

This commit updates the `tests/docker/environment.yml` so that it
includes `sphinx_autodoc_typehints` as well as `mypy`.


src/stk/molecular/atoms/atom.py
```diff
@@ -15,103 +17,107 @@ class Atom:
...
+    _atomic_number: ClassVar[int]
 
-    def __init_subclass__(cls, **kwargs):
+    def __init_subclass__(cls: Type['Atom'], **kwargs) -> None:
         # Replace the default __init__() method of the subclass with
         # _subclass_init(). This is because the default __init__()
         # method takes an atomic_number parameter, but
         # _subclass_init() does not.
-        cls.__init__ = cls._subclass_init
+        cls.__init__ = cls._subclass_init  # type: ignore
```
andrewtarzia, line 64: Why is this type: ignore?

src/stk/molecular/atoms/atom.py
```diff
@@ -15,103 +17,107 @@ class Atom:
...
 
     """
 
     # Maps each atomic number (int) to the relevant Atom subclass.
-    _elements = {}
+    _elements: Dict[int, Type['Atom']] = {}
+    _atomic_number: ClassVar[int]
 
-    def __init_subclass__(cls, **kwargs):
+    def __init_subclass__(cls: Type['Atom'], **kwargs) -> None:
```
andrewtarzia, line 59: Why not cls: Atom ?

andrewtarzia (approved):

src/stk/molecular/atoms/atom.py
lukasturcani: Because `cls; Atom` is saying `cls` is an in instance of `Atom` while `Type['Atom']` is saying `cls` is a subclass of `Atom`

src/stk/molecular/atoms/atom.py
lukasturcani: because I'm chaning the signature of the function, and therefore it's type. Mypy, quite rightly, complains about a type mismatch. However, in this case I think it's fine. The alternative would be to specify the innit function for each element, which is a lot of duplication, or add a new "implmentation" atom class, which the  element classes inherit, which has the init method defined. that would work pretty well, but i wanted to keep the inheritence hierarchy shallow.

andrewtarzia: Cool cool, that makes sense. So this is a note for mypy basically?

src/stk/molecular/atoms/atom.py
andrewtarzia: Did not know. Thank you!

src/stk/molecular/atoms/atom.py
lukasturcani: yeah. to be honest though. thinking about it now, i think the current way of replacing the init method is more complex and less obvious than just using simple inheritance. So i think I will just make the AtomImpl class.

Issue #349, opened 2021-06-16
replace use of write() in examples with Molecular writers


lukasturcani:
The `Molecule.write()` method is for convenience and I want it to be obvious that the correct way to write molecules to files is with writers.

PR #363, opened 2021-07-08
User Writer class in docs.

andrewtarzia:
Related Issues: #349 

This PR simply replaces examples of `Molecule.write(path)` with `stk.Writer().write(molecule, path)` in prominent places that do not affect convenience (e.g. in the EA runs, some places should stay in the original version because of actual convenience and unclear file type expectations).


lukasturcani:
Amazing ty!

lukasturcani (approved):

Issue #366, opened 2021-07-16
Using same position matrix collection but distinct molecule collections produces error.


andrewtarzia:
This issue was first noticed by @annabelbasford.

Basically, if you use different molecule collections but the same pos_mat_collection in a molecule database, the .get_all method will hit an error because it seems to run through the pos_mat collection in order.

A minimum working example to outline it:

```python
import stk
import pymongo

molecule1 = stk.BuildingBlock('NCCN')
molecule2 = stk.BuildingBlock('CN')
molecule3 = stk.BuildingBlock('CC')
molecule4 = stk.BuildingBlock('NCCCCN')

client = pymongo.MongoClient()
db = stk.MoleculeMongoDb(
    mongo_client=client, 
    database='random', 
    molecule_collection='random2', 
    position_matrix_collection='constant',
)

db.put(molecule1)
db.put(molecule2)
for entry in db.get_all():
    print(stk.BuildingBlock.init_from_molecule(entry))
# Works up to here!

db2 = stk.MoleculeMongoDb(
    mongo_client=client, 
    database='random', 
    molecule_collection='random3', 
    position_matrix_collection='constant',
)

db2.put(molecule3)
db2.put(molecule4)

for entry in db.get_all():
    print(stk.BuildingBlock.init_from_molecule(entry))
    # Prints first two, then `KeyError`
for entry in db2.get_all():
    print(stk.BuildingBlock.init_from_molecule(entry))
    # `KeyError`
```

andrewtarzia:
Discussion point:

The current MoleculeMongoDB.get_all() method looks like this:

```python
    def get_all(self):
        for entry in self._position_matrices.find():
            # Do 'or' query over all key value pairs.
            query = {'$or': [
                {key: value}
                for key, value in self._get_molecule_keys(entry)
            ]}

            json = self._molecules.find_one(query)
            if json is None:
                raise KeyError(
                    'No molecule found in the database associated '
                    f'with a position matrix with query: {query}. '
                    'This suggests your database is corrupted.'
                )

            yield self._dejsonizer.from_json({
                'molecule': json,
                'matrix': entry,
            })

```

I would like to discuss why we think that should be an error there, instead of just passing the case where a molecule does not have a position matrix, or vice-versa? There is no obligation that they are both there for any given entry. I say this one should just `continue` to the next entry.

andrewtarzia:
[...]

PR #368, opened 2021-07-17
Fix `get_all` for mongo dbs.

andrewtarzia:
Related Issues: #366 

There are no changes to the interface. These changes just ensure that `.get_all()` methods appropriately filter for entries that exist in all relevant collections of the database. However, the user will be a speedup! The speedup is the reason behind not just using `continue` -- see the associated issue for discussion and below for benchmarks.

The main risk here, and why this is a draft, is that I am not a MongoDB expert, so I think this could be made more robust/cleaner/rigorous in ensuring only one entry is found in all cases. I expect this review to find these issues and improve them. Another issue is that the implementation explicitly relies upon the scheme (i.e. `a`, `b` and `m` for atoms, bonds and pos mat) to be correct.

* `src/stk/databases/mongo_db/constructed_molecule.py:ConstructedMoleculeMongoDb.get_all()`: The query was updated.
* `src/stk/databases/mongo_db/molecule.py:MoleculeMongoDb.get_all()`: The query was updated.

An example code that uses a small database to test for the error in the issue, and then a larger (my own database) to benchmark is below.

Timings (see print statements in code below, but `eg4` is the iteration over my 500 entry long database):
With `continue`:
eg1 0.007266521453857422
eg2 0.00848388671875
eg3 0.011469125747680664
**eg4 53.28781604766846**

With new query:
eg1 0.008028030395507812
eg2 0.006296873092651367
eg3 0.011692047119140625
**eg4 3.042726755142212**

```python
import stk
import pymongo
import time

molecule1 = stk.BuildingBlock('NCCN')
molecule2 = stk.BuildingBlock('CN')
molecule3 = stk.BuildingBlock('CC')
molecule4 = stk.BuildingBlock('NCCCCN')
molecule5 = stk.BuildingBlock('C')
molecule6 = stk.BuildingBlock('CCC')
molecule7 = stk.BuildingBlock('CCCC')
molecule8 = stk.BuildingBlock('CCCCC')
molecule9 = stk.BuildingBlock('CCCCCC')
molecule10 = stk.BuildingBlock('CCCCCCC')


client = pymongo.MongoClient()
db = stk.MoleculeMongoDb(
    mongo_client=client,
    database='random',
    molecule_collection='random2',
    position_matrix_collection='constant',
)

db.put(molecule1)
db.put(molecule2)
db.put(molecule5)
db.put(molecule6)
db.put(molecule7)
db.put(molecule8)
db.put(molecule9)
db.put(molecule10)
start = time.time()
for entry in db.get_all():
    a = stk.BuildingBlock.init_from_molecule(entry)
print('eg1', time.time() - start)
# Works up to here!
db2 = stk.MoleculeMongoDb(
    mongo_client=client,
    database='random',
    molecule_collection='random3',
    position_matrix_collection='constant',
)

db2.put(molecule3)
db2.put(molecule4)

start = time.time()
for entry in db.get_all():
    a = stk.BuildingBlock.init_from_molecule(entry)
    # Prints first two, then `KeyError`

print('eg2', time.time() - start)
start = time.time()
for entry in db2.get_all():
    a = stk.BuildingBlock.init_from_molecule(entry)
    # `KeyError`
print('eg3', time.time() - start)

# Benchmarking with larger db:
user = 'testuser'
password = 'testpw'
client = pymongo.MongoClient(
    f'mongodb+srv://{user}:{password}@cluster0.32as5.mongodb.net'
    '/lc?retryWrites=true&w=majority'
)
jsonizer = stk.MoleculeJsonizer((stk.Smiles(), ))
indices = (stk.Smiles().get_key_name(), )
pd_db = stk.MoleculeMongoDb(
    mongo_client=client,
    database='lc_molecules',
    molecule_collection='built_molecules',
    position_matrix_collection='opt_chosen',
    jsonizer=jsonizer,
    indices=indices
)
start = time.time()
for entry in pd_db.get_all():
    a = stk.BuildingBlock.init_from_molecule(entry)
print('eg4', time.time() - start)
```


andrewtarzia:
Looking at the failing tests I see an issue. The current approach uses the indices to define the keys to collect entries from - but a key used to `put` does not have to have anything to do with the indices, so the tests fail that use the `inchi` key maker.

There is no clear way to get all relevant keys in a collection, and ensuring that duplicates are not found - that I can see. 

Would it be impossible to request a `keys` argument in the `get_all` methods? This would solve the issue completely and works equivalently to how stk-vis might handle this issue.

lukasturcani:
I think there's 2 ways to resolve this. 

1. Make the tests use indices correctly. Really if you're putting an molecule with a certain key, you should have made an index on that key. (Note this only needs to be done once, not every time the `MoleculeMongoDb` instance is created.
2. First use `find_one`, to get an entry in the molecules collection, and then check which keys it holds.

I have a pretty strong preference for option 1.

andrewtarzia:
Ok, I agree with that statement but, as discussed, this requires changes to the interface that allows `indices` to be ignored.

See the new changes.

Unfortunately, the test still fails. But I cannot see why that db would be empty - any thoughts?

src/stk/databases/mongo_db/constructed_molecule.py
```diff
@@ -689,38 +689,86 @@ def _get_molecule_keys(self, entry):
...
-                        self._get_building_block,
-                        entry['BB'],
-                    ))
-                },
+        keys = tuple(set((
+            val['key'][0][0]
+            for val in (
+                list(pos_mat_indices.values())
+                + list(molecules_indices.values())
+                + list(constmolecules_indices.values())
```
lukasturcani, line 704: how bout

```python
indices = itertools.chain(
    self._positoin_matrices.index_information().values(),
    self.molecules.index_informatoin.values(),
    self._constructed_molecules.index_information.values(),
)
keys = dedupe(
   index['key'][0][0] for index in indices 
)
```

* Uses chain instead of list concatenation - O(1) vs O(N) memory use, + O(N) vs O(N^2) time
* No need to make a tuple at the end, since you are only using keys to iterate though them: O(1) vs O(N) memory + O(1) vs O(N) time.

lukasturcani (changes requested):

src/stk/databases/mongo_db/constructed_molecule.py
andrewtarzia: I like it!

andrewtarzia:
Also solves #369!

src/stk/databases/mongo_db/constructed_molecule.py
```diff
@@ -689,38 +691,78 @@ def _get_molecule_keys(self, entry):
...
+                        'from': self._position_matrices.name,
+                        'localField': key,
+                        'foreignField': key,
+                        'as': 'posmat',
+                    },
+                },
+                {
+                    '$match': {
+                        '$expr': {
+                            '$or': [
```
lukasturcani, line 732: I think this should be "and" not "or"? cos both the mol and the posmat need to exist.

src/stk/databases/mongo_db/constructed_molecule.py
```diff
@@ -689,38 +691,78 @@ def _get_molecule_keys(self, entry):
...
+        indices = chain(
+            self._position_matrices.index_information().values(),
+            self._molecules.index_information().values(),
+            self._constructed_molecules.index_information().values(),
+        )
+        keys = dedupe(index['key'][0][0] for index in indices)
+
+        # Iterate over potential keys, and aggregate matching position
+        # matrices with molecules and constructedmolecules.
+        for key in keys:
```
lukasturcani, line 704: if we go through all the keys, we can end up with yield each molecule twice, which is different to the current behaviour and potentially undesirable. For example , if we used both the inchikey and smiles, we would yield each molecule twice, once for inchikey and once for smiles. The simple solution is to have set which keeps track of yielded molecules, but this is not a reasonable solution because it takes O(N) memory, meaning using `get_all` would become impossible for sufficiently large databases.

I think what needs to happen is that we need to programatically create the query passed to the `aggregate` method. Meaning that `$lookup` dictionary is inserted into the query for each key. Similarly a `$match` query will be added too. For example 

```python

# need to make a tuple cos keys is gonna be reused
keys = tuple(dedupe(...))
query = [
    {
        '$match': {
            '$expr': {
                  '$or': [
                           {key: {'$exists': True}}
                           for key  in keys
                   ], 
             }, 
        }
    },
]
query.extend(
    {
        '$lookup': {  
              'from': self._molecules.name,
              'localField': key,
              'foreignField': key,
              'as':f 'mol_{key}',
        },
    }
    for key in keys
)
# do lookups for posiation matrix and stuff in the same way
query.append(
    {   
         '$match': {
                 '$expr': {
                          '$or': [
                                      {
                                         '$and': [
                                               {
                                                  '$gt': [
                                                      {'$size': f'$mol_{key}'},
                                                      0
                                                  ],
                                              },
                                              {
                                                  '$gt': [
                                                      {'$size': f'$posmat_{key}'},
                                                      0
                                                  ],
                                              },                                    
                                          ],
                                      }
                                     for key in keys
                            ]
                  },
         },
    }
)

```
(disclaimer - I'm not sure if I've botched this code- my main worry is the `$and` nested in the `$or` - it is for illustrative purposes only)

I think a test case making sure that we get the correct behaviour for a database that has the following entries:
* some molecules with 2 different molecule keys,
* some molecules with only one of the molecule keys
* The database has an index made for both of the molecule keys

lukasturcani (changes requested):

andrewtarzia:
Some updated timings (same test script as above):

eg1 0.009424448013305664
eg2 0.008942842483520508
eg3 0.009641647338867188
672 molecules found.
eg4 5.066543340682983

tests/databases/constructed_molecule/get_all/fixtures/constructed_molecule_mongo_dbs.py
```diff
@@ -0,0 +1,116 @@
...
+                    building_blocks[4], building_blocks[5]
+                ),
+                repeating_unit='AB',
+                num_repeating_units=3,
+            ),
+        ),
+    ]
+
+
+def get_database(database_name, mongo_client, keys, indices):
```
lukasturcani, line 70: type hints

tests/databases/constructed_molecule/get_all/fixtures/constructed_molecule_mongo_dbs.py
```diff
@@ -0,0 +1,116 @@
...
+        database=database_name,
+        jsonizer=stk.ConstructedMoleculeJsonizer(keys),
+        put_lru_cache_size=0,
+        get_lru_cache_size=0,
+        indices=indices,
+    )
+
+
+@pytest.fixture
+def constructed_molecule_mongo_dbs(mongo_client, molecules):
```
lukasturcani, line 83: type hints

tests/databases/constructed_molecule/get_all/test_get_all.py
```diff
@@ -0,0 +1,97 @@
...
+        The database to test collection from.
+
+    molecules : :class:`iterable` of :class:`.Molecule`
+        The molecules to put and get from the databases.
+
+    Returns
+    -------
+    None : :class:`NoneType`
+
+    """
```
lukasturcani, line 62: This test needs to be reworked. There are 2 issues here

* Not every database will use inchi and smiles, or molecule keys at all. Having variables refer to them is leaking something specific to one database -> the test is not generic enough.
* All the putting of stuff into the a database is not the actual test, it is test setup. it should therefore be part of a fixture, not part of the test itself.

lukasturcani: I think the test should be, something along the lines of 


```python

def test_get_all(case_data: CaseData) -> None:
    _test_get_all(
        database=case_data.database,
        expected_molecules=case_data.expected_molecules
    )

def _test_get_all(...) -> None:
        smiles = stk.Smiles()
        for i, retrieved in enumerate(database.get_all()):
            expected = expected_molecules[smiles.get_key(retrieved)]
            is_equivalent_molecule(...)
        
        # the rest
```

THe only pieces of data the test-case should contain is the database we're testing and a map mapping the smiles to the expected molecule

lukasturcani (changes requested):

src/stk/databases/mongo_db/constructed_molecule.py
```diff
@@ -689,38 +691,146 @@ def _get_molecule_keys(self, entry):
...
-            position_matrix = self._position_matrices.find_one(query)
-            if position_matrix is None:
-                raise KeyError(
-                    'No position matrix found in the database '
-                    'associated with a position matrix with query: '
-                    f'{query}. This suggests your database is '
-                    'corrupted.'
-                )
+        # Get all potential indices.
+        indices = chain(
```
lukasturcani, line 695: `itertool.chain`

lukasturcani (approved):

PR #372, opened 2021-07-27
Use UnaligningVertex in LinearPolymer class.

andrewtarzia:
Related Issues:  #156

This PR makes no changes visible to the user but allows the user to placer molecules like `BrCBr` on the polymer.linear topology graph using UnaligningVertex. This is handled by the Linear initializer when defining vertices (all vertices with building blocks that only have one placer are converted to a UnaligningVertex - same as in Cof).

Added a new init class method to the UnaligningVertex to allow this as the __init__ class is used elsewhere and did not want to change that - happy to discusss.

* `src/stk/molecular/topology_graphs/polymer/linear/linear.py:_get_building_block_vertices`: Added a check for whether each vertex should be an UnaligningVertex.

* `src/stk/molecular/topology_graphs/polymer/linear/vertices.py:UnaligningVertex`: Added new init class method.

Two new tests added for vertex and molecule fixture.


src/stk/molecular/topology_graphs/polymer/linear/linear.py
```diff
@@ -440,8 +440,27 @@ def _get_building_block_vertices(self, building_blocks, vertices):
...
             building_block_vertices[bb].append(vertex)
+
+        building_block_vertices = self._with_unaligning_vertices(
+            building_block_vertices=building_block_vertices,
+        )
+
         return building_block_vertices
 
+    @staticmethod
+    def _with_unaligning_vertices(building_block_vertices):
```
lukasturcani, line 451: needs type hints

src/stk/molecular/topology_graphs/polymer/linear/linear.py
```diff
@@ -440,8 +440,27 @@ def _get_building_block_vertices(self, building_blocks, vertices):
...
+        clone = dict(building_block_vertices)
+        for building_block, vertices in clone.items():
+            # Building blocks with 1 placer, cannot be aligned and
+            # must therefore use an UnaligningVertex.
+            if len(set(building_block.get_placer_ids())) == 1:
```
lukasturcani, line 456: Looks like we need to add a num_placer_ids method. (#376)

src/stk/molecular/topology_graphs/polymer/linear/vertices.py
```diff
@@ -196,6 +196,21 @@ class UnaligningVertex(LinearVertex):
 
     """
 
+    @classmethod
+    def init_from_vertex(cls, vertex):
+        """
+        Initialize from a :class:`LinearVertex`.
+        Parameters
```
lukasturcani, line 203: Missing newline and needs to use Google-style comments.

src/stk/molecular/topology_graphs/polymer/linear/vertices.py
```diff
@@ -196,6 +196,21 @@ class UnaligningVertex(LinearVertex):
 
     """
 
+    @classmethod
+    def init_from_vertex(cls, vertex):
+        """
+        Initialize from a :class:`LinearVertex`.
```
lukasturcani, line 202: missing dot: 
```
:class:`.LinearVertex`
```

src/stk/molecular/topology_graphs/polymer/linear/vertices.py
```diff
@@ -196,6 +196,21 @@ class UnaligningVertex(LinearVertex):
 
     """
 
+    @classmethod
+    def init_from_vertex(cls, vertex):
```
lukasturcani, line 200: Missing type hints.

You will also need to add a 
```python
from __future__ import annotations
```
at the top of this file, because the return type of the init method is the same type as the class, which requires the future import.

src/stk/molecular/topology_graphs/polymer/linear/linear.py
```diff
@@ -440,8 +440,27 @@ def _get_building_block_vertices(self, building_blocks, vertices):
...
+                clone[building_block] = tuple(map(
+                    UnaligningVertex.init_from_vertex,
```
lukasturcani, line 458: I would just do 
```python
tuple(
  UnaligningVertex(
      id=vertex.get_id(),
      position=vertex.get_position(),
      flip=vertex.get_flip(),
  ) for vertex in vertices
)
```

and remove the new `init` method. I think methods should be added fairly rarely and only when they are needed. I think it is not needed in this case.

lukasturcani (changes requested):
My only reservation is that some type hints are missing and that we don't need a new init method.

src/stk/molecular/topology_graphs/polymer/linear/linear.py
```diff
@@ -4,10 +4,13 @@
 
 """
 
+from __future__ import annotations
+
 from dataclasses import dataclass
+from typing import Iterable, Tuple
```
lukasturcani, line 10: No need to use `Tuple` use the built-in `tuple` instead.

src/stk/molecular/topology_graphs/polymer/linear/linear.py
```diff
@@ -431,7 +434,11 @@ def _normalize_repeating_unit(repeating_unit):
         base = ord('A')
         return tuple(ord(letter)-base for letter in repeating_unit)
 
-    def _get_building_block_vertices(self, building_blocks, vertices):
+    def _get_building_block_vertices(
+        self,
+        building_blocks: dict[int, BuildingBlock],
+        vertices: Iterable[Vertex],
+    ) -> dict[BuildingBlock, type[Vertex]]:
```
lukasturcani, line 441: this is wrong, it maps a building block to a vertex instance,  not building block to the Vertex class

```python
dict[BuildingBlock, Vertex]
```

tests/molecular/topology_graphs/vertex/cases/linear/conftest.py
```diff
@@ -47,6 +47,16 @@ def tail(id, position, flip):
     )
 
 
+@pytest.fixture
+def unaligning(id, position, flip):
```
lukasturcani, line 51: missing type hints

lukasturcani (changes requested):

andrewtarzia:
Closing to be implemented from scratch and cleaned.

Issue #370, opened 2021-07-27
Improve test startup time


lukasturcani:
Currently lots of tests create test cases by making instances in the following way

```python

@fixture(
    params=(
        Something(),
    ),
)
def some_fixture(request):
    return request.param

```

this is a problem for two reasons:

* if `Something()` breaks, it breaks before any tests are done and the pytest output doesn't actually tell us much about the error
* There is a long test collection time because all the test cases are getting initialized as part of test collection. This makes it really slow run pyest on specific tests, because the collection time takes up so much time

Solution, defer creation of fixture instances until after test collection

```python

@fixture(
    params=(
        lambda: Something(),
    ),
)
def some_fixture(request):
    return request.param()

```

lukasturcani:
Note to self: in the upcoming PR, remember  to check that the total number of tests is the same and that if you've added fixtures, that there are no name conflicts between the added fixtures. Also check if there is a way to report if there is a fixture name conflict.

PR #375, opened 2021-07-28
Improve test collection speed

lukasturcani:
Related Issues: #370 

The speed at which tests are collected has been dramatically improved. Previously, running
```bash
pytest --collect-only
```

would take approximately 12.84 s. It is now 3.07 s. This is extremely important because collection time is spent every time pytest is run, even if only a subset of tests are run. For example

```bash
pytest -k get_id
```

would have spent 12.84 s collecting all the tests and then about 1 s to actually run the tests. This is an incredibly frustrating developer experience because it makes running tests very unresponsive. Now running the same command would take 4.07 s.
Importantly the total time spent running tests is basically unchanged, going from 48.24 s previously to 48.78 s now.

Another important change is that test will no longer break in the collection phase, or at least, very few of them will. The issue with tests breaking during the collection phase is that if this happens, pytest does not provide very good error information.

Both the collection time and prevention of breakage during the collection change was achieved by deferring the creation of test cases to after the collection phase. For example, a fixture such as this

```python
@pytest.fixture(
    params=(
        CaseData(...),
        CaseData(...),
    ),
)
def case_data(request):
    return request.param

```

has been changed to this

```python
@pytest.fixture(
    params=(
        lambda: CaseData(...),
        lambda: CaseData(...)
    ),
)
def case_data(request):
    return request.param()
```

The difference between these two pieces of code is that in the original version the `CaseData` instance is created when the module is being read. However, in the new version, the `CaseData` instance is created only once the fixture is being used by the test. This means the collection phase of pytest does not actually create any instances any more, the test execution phase creates them now.

The other oft-repeated change is turning module-level variables into functions

```python
some_var = Something()
```

into functions

```python
def _get_some_var() -> Something:
    return Seomthing()
```

Again, this means that instances are not created when the module is being read. This change however does mean that the variable can no longer be re-used as a new instance will be created every time `_get_some_var()` is called. For most cases, I expect this will make no practical difference. In cases where I did think a performance impact may be incurred I used fixtures

```python
@pytest.fixture(scope='session')
def _get_some_var() -> Something:
    return Something()
```

These changes constitute the vast majority of the changes, some other notable changes are:
* `pytest.ini`: Added this file and explicitly specified the directory where the tests are located. This means that running `pytest` will only search the `tests` directory rather than all other other directories too when looking for tests, giving a small reduction in test collection time.
* `tests/molecular/molecules/.../fixtures/building_blocks.py`: I added this file to factor out the common building blocks used across multiple modules.


andrewtarzia:
My only comment is that you should delete all .npy and rebuild all (git commit should only be delete 2, add 2), because I think you added 2 instead of replacing 2.

andrewtarzia (approved):

lukasturcani:
> My only comment is that you should delete all .npy and rebuild all (git commit should only be delete 2, add 2), because I think you added 2 instead of replacing 2.

Could you spell it out a little more for me please?

andrewtarzia:
You changed two test names in the cage fixtures, which lead to two new .npy files. But you did not delete the .npy files from the old name. Right?

lukasturcani:
I did, that's why the the files are listed as renamed.

andrewtarzia:
Oops misread!

Issue #376, opened 2021-07-29
Add get_num_placer_ids()


lukasturcani:
Would be useful for #372

PR #377, opened 2021-07-29
Add BuildingBlock.get_num_placers()

lukasturcani:
Related Issues: #376

A new method, `BuildingBlock.get_num_placers()` has been added. This
is necessary because sometimes the number of placer atoms is necessary,
for example, by unaligning vertices, which needs to check the number of
placer atoms, in order to determine if the unaligning vertex should be
used. Previously all the placer ids would have to be iterated though
with `BuildingBlock.get_placer_ids()`, which is both ugly and less
performant.


andrewtarzia (approved):
Perfect

Issue #276, opened 2021-01-04
placer ids in buildingblock need to be deduped when collected from functional groups.


andrewtarzia:
This is a simple dedupe in Molecule class, _init_from_rdkit_mol (I think).

PR #390, opened 2021-08-31
Use set for placer ids.

andrewtarzia:
Related Issues: #372 , #276, #156

Found, while working on #372, that BuildingBlock._placer_ids contained duplicate ids from distinct functional groups. This PR fixes this issue by using a frozen set in BuildingBlock.py.

Main change in 
* `src/stk/molecular/molecules/building_block.py:BuildingBlock._normalize_placer_ids()`: This method now returns a set.

Added a test of this specific issue.

Also added the `BuildingBlock._core_ids` attribute and the `BuildingBlock._get_core_ids()` method such that this property, similar to `_placer_ids` is set on initialization of the `BuildingBlock` instead of collected from scratch on call of `BuildingBlock.get_core_atom_ids()`.

Many position matrices change negligibly but enough to fail the tests due to new centroid positions due to the set.  @lukasturcani please visualise structures to see that things look about right to you?


src/stk/molecular/molecules/building_block.py
```diff
@@ -644,15 +653,50 @@ def _normalize_placer_ids(self, placer_ids, functional_groups):
...
+        Get the final *core* ids.
+
+        Parameters:
+
+            functional_groups:  The :class:`.FunctionalGroup` instances
+            of the building block.
+
+        Returns:
+
+            core_ids: Ids of atoms defining the core.
```
lukasturcani, line 680: no label here plz

src/stk/molecular/molecules/building_block.py
```diff
@@ -644,15 +653,50 @@ def _normalize_placer_ids(self, placer_ids, functional_groups):
...
+        self,
+        functional_groups: Iterable[FunctionalGroup],
+    ) -> frozenset[int]:
+        """
+        Get the final *core* ids.
+
+        Parameters:
+
+            functional_groups:  The :class:`.FunctionalGroup` instances
+            of the building block.
```
lukasturcani, line 676: indent. the second line. You can also indent like this

```
functional_groups:
    The :class:`.FunctionalGroup` instances of 
    the building block.
```

src/stk/molecular/molecules/building_block.py
```diff
@@ -644,15 +653,50 @@ def _normalize_placer_ids(self, placer_ids, functional_groups):
...
+
+        Returns:
+
+            core_ids: Ids of atoms defining the core.
+
+        """
+
+        core_ids = []
```
lukasturcani, line 684: If you make this a set, rather than a list, your memory requirements while this function is running will be lower.

src/stk/molecular/molecules/building_block.py
```diff
@@ -644,15 +653,50 @@ def _normalize_placer_ids(self, placer_ids, functional_groups):
...
                 for functional_group in functional_groups
             ))
 
-        return tuple(atom.get_id() for atom in self._atoms)
+        return frozenset(atom.get_id() for atom in self._atoms)
+
+    def _get_core_ids(
+        self,
+        functional_groups: Iterable[FunctionalGroup],
+    ) -> frozenset[int]:
```
lukasturcani, line 669: Tbh, maybe making this a generator would be better, where the generator will return duplicates. My reasoning being that you can then call it like this

```python
self._core_ids = frozenset(self._get_core_ids(...))
```

Where its immediately obvious that you re deduping.  With

```python
self._core_ids = self._get_core_ids(...)
```

The reader has to jump to the function definition to see the frozenset.

You'll have to add a comment saying that it might return dupes though.

src/stk/molecular/molecules/building_block.py
```diff
@@ -610,24 +615,28 @@ def _init_from_rdkit_mol(
...
-        Parameters
-        ----------
-        placer_ids : :class:`tuple` of :class:`int`
-            The ids of *placer* atoms or ``None``.
+        Parameters:
 
-        functional_groups : :class:`iterable`
-            The :class:`.FunctionalGroup` instances of the building
-            block.
+            placer_ids: The ids of *placer* atoms or ``None``.
```
lukasturcani, line 632: If placer ids can be then, then the type hint is wrong, and should be `Optional[tuple[int, ...]]`

src/stk/molecular/molecules/building_block.py
```diff
@@ -644,15 +653,50 @@ def _normalize_placer_ids(self, placer_ids, functional_groups):
         """
 
         if placer_ids is not None:
-            return placer_ids
+            return frozenset(placer_ids)
```
lukasturcani, line 656: If we require the type signature to take a `frozenset` instead of a `tuple` we do not have to do the conversion here.

src/stk/molecular/topology_graphs/cage/vertices.py
```diff
@@ -290,14 +290,6 @@ class UnaligningVertex(_CageVertex):
 
     """
 
-    def __init__(self, vertex):
```
lukasturcani, line 293: Why is this being removed?

src/stk/molecular/topology_graphs/cof/vertices.py
```diff
@@ -310,14 +310,6 @@ class UnaligningVertex(_CofVertex):
 
     """
 
-    def __init__(self, vertex):
```
lukasturcani, line 313: Why is this being removed?

lukasturcani (changes requested):
Pretty close.

lukasturcani:
An explanation of why the `_core_atom_ids` attribute needed to be created would be good in the PR message.

src/stk/molecular/topology_graphs/cage/vertices.py
andrewtarzia: It was a previous request for handling unaligning vertices properly to fix all the init methods to not take the vertex, but instead to be a normal vertex init method.

src/stk/molecular/topology_graphs/cof/vertices.py
andrewtarzia: It was a previous request for handling unaligning vertices properly to fix all the init methods to not take the vertex, but instead to be a normal vertex init method.

src/stk/molecular/molecules/building_block.py
lukasturcani: Hehe, check again. 
hint: ...

src/stk/molecular/molecules/building_block.py
```diff
@@ -610,24 +615,28 @@ def _init_from_rdkit_mol(
...
-            The :class:`.FunctionalGroup` instances of the building
-            block.
+            placer_ids: The ids of *placer* atoms or ``None``.
 
-        Returns
-        -------
-        :class:`tuple` of :class:`int`
-            Depending on the input  values, this function will return
+            functional_groups:  The :class:`.FunctionalGroup` instances
+            of the building block.
```
lukasturcani, line 635: indent missing

src/stk/molecular/molecules/building_block.py
```diff
@@ -644,15 +653,53 @@ def _normalize_placer_ids(self, placer_ids, functional_groups):
         """
 
         if placer_ids is not None:
-            return placer_ids
+            return frozenset(placer_ids)
 
         if functional_groups:
-            return tuple(flatten(
+            return frozenset(flatten(
```
lukasturcani, line 659:
```python
if functional_groups
```

won't work if `functional_groups` is an Iterable.

src/stk/molecular/molecules/building_block.py
```diff
@@ -644,15 +653,53 @@ def _normalize_placer_ids(self, placer_ids, functional_groups):
...
+        return frozenset(atom.get_id() for atom in self._atoms)
+
+    def _get_core_ids(
+        self,
+        functional_groups: Iterable[FunctionalGroup],
+    ) -> Iterator[int]:
+        """
+        Get the final *core* ids.
+
+        This method may return duplicates.
```
lukasturcani, line 673: This method may return duplicate ids.

src/stk/molecular/molecules/building_block.py
```diff
@@ -644,15 +653,53 @@ def _normalize_placer_ids(self, placer_ids, functional_groups):
...
+
+        Parameters:
+
+            functional_groups:
+                The :class:`.FunctionalGroup` instances of the building
+                block.
+
+        Yields:
+
+            Ids of atoms defining the core.
```
lukasturcani, line 683: Yields comments are in the singlar, they describe a yielded item:  

Id of an atom defining the core of the molecule.

src/stk/molecular/topology_graphs/cage/vertices.py
lukasturcani: by previous request, do you mean issue?

lukasturcani (changes requested):
closer.

src/stk/molecular/molecules/building_block.py
andrewtarzia: Ahh forgot to change this back after fixing it and breaking it.

src/stk/molecular/topology_graphs/cage/vertices.py
andrewtarzia: From previous times we discussed this issue (and the other related issues) we went for wanting a normal init method for all vertices. I figured this change is small and fits here.

src/stk/molecular/molecules/building_block.py
andrewtarzia: Are you thinking `if len(functional_groups) > 0`:

lukasturcani: That won't work with an iterable.

src/stk/molecular/topology_graphs/cage/vertices.py
lukasturcani: why wasn't it using the normal init method?

andrewtarzia: No idea - a previous implentation by one of us to use `UnaliningVertex(vertex)` and then call `super.__init__(...)` - which we now just do `UnaliningVertex(...)`

lukasturcani: I guess it was just a convenience thing?

tests/molecular/topology_graphs/vertex/cases/cof/fixtures/cof1.py
```diff
@@ -23,12 +23,10 @@ def cof1(cls, id, position, aligner_edge, cell):
...
-            vertex=cls(
-                id=id,
-                position=position,
-                aligner_edge=aligner_edge,
-                cell=cell,
-            )
+            id=id,
+            position=position,
+            aligner_edge=aligner_edge,
+            cell=cell,
```
lukasturcani, line 29: here the cls fixture is no longer being used here `unaliged_cof` fixutre -> I assume we can remove it?

lukasturcani (changes requested):

tests/molecular/topology_graphs/vertex/cases/cof/fixtures/cof1.py
lukasturcani: can we remove this whole fixture?

lukasturcani (approved):

PR #393, opened 2021-09-01
Use unaligning vertex in Linear polymer

andrewtarzia:
Related Issues: #156

This PR makes no changes visible to the user but allows the user to place molecules like `BrCBr` on the `polymer.linear` topology graph using `UnaligningVertex`. This is handled by the Linear initializer when defining vertices (all vertices with building blocks that only have one placer are converted to a UnaligningVertex - same as in Cof and Cage).

* `src/stk/molecular/topology_graphs/polymer/linear/linear.py:_get_building_block_vertices`: Added a check for whether each vertex should be an UnaligningVertex.

* `src/stk/molecular/topology_graphs/polymer/linear/vertices.py:UnaligningVertex`: Added new init class method.

Two new tests added for vertex and molecule fixture.


src/stk/molecular/topology_graphs/polymer/linear/linear.py
```diff
@@ -4,10 +4,13 @@
 
 """
 
+from __future__ import annotations
+
 from dataclasses import dataclass
+from typing import Iterable, Tuple
```
lukasturcani, line 10: do not use `typing.Tuple` using `tuple`  (that's why the `__future__` import is there)

tests/molecular/topology_graphs/vertex/cases/linear/conftest.py
```diff
@@ -47,6 +47,20 @@ def tail(id, position, flip):
     )
 
 
+@pytest.fixture
+def unaligning(id, position, flip):
```
lukasturcani, line 51: thats a nice fixture you have there. is it being used anywhere?

src/stk/molecular/topology_graphs/polymer/linear/linear.py
```diff
@@ -578,7 +581,11 @@ def _normalize_repeating_unit(repeating_unit):
         base = ord('A')
         return tuple(ord(letter)-base for letter in repeating_unit)
 
-    def _get_building_block_vertices(self, building_blocks, vertices):
+    def _get_building_block_vertices(
+        self,
+        building_blocks: dict[int, BuildingBlock],
+        vertices: Iterable[Vertex],
+    ) -> dict[BuildingBlock, type[Vertex]]:
```
lukasturcani, line 588: the

```python
type[Vertex]
```

annotation is wrong I think, what are you trying to express?

src/stk/molecular/topology_graphs/polymer/linear/linear.py
```diff
@@ -587,9 +594,35 @@ def _get_building_block_vertices(self, building_blocks, vertices):
...
+            building_block_vertices=building_block_vertices,
+        )
         return building_block_vertices
 
-    def _get_scale(self, building_block_vertices):
+    @staticmethod
+    def _with_unaligning_vertices(
+        building_block_vertices: dict[BuildingBlock, type[Vertex]],
+    ) -> dict[BuildingBlock, type[Vertex]]:
+        clone = dict(building_block_vertices)
```
lukasturcani, line 607: you don't haveto clone it, just make an empty dict and then fill it in the for loop

lukasturcani (changes requested):
close

src/stk/molecular/topology_graphs/polymer/linear/linear.py
andrewtarzia: That was you ;)

src/stk/molecular/topology_graphs/polymer/linear/linear.py
andrewtarzia: Correct, is wrong. It is a Vertex.

src/stk/molecular/topology_graphs/polymer/linear/linear.py
lukasturcani: to be fair, that was before i was doing the future imports

lukasturcani (approved):

PR #402, opened 2021-10-02
Unqualify imports

lukasturcani:
Stop using qualified imports. Qualified imports can mean that failed
imported are silently ignored in type hints.


andrewtarzia (commented):
I love this change.

PR #405, opened 2021-10-18
Add type hints

lukasturcani:


lukasturcani:
Argh, this PR is too big, and I don't need to even restructure the files.... gonna start from scratch

PR #408, opened 2021-11-14
Revert type hints

lukasturcani:
This PR reverts a bunch of commits related to type hints.

The issue with the type hint commits, is that I did not realize I could
avoid changing the file structure of `stk`. So this commit will revert
all the changes and I will start making new type hint commits, which
do not restructure the folders of  `stk`.


andrewtarzia:
Frustrating, but glad to hear it to be fair.

Issue #446, opened 2022-05-12
Remove `pdb` from init_from_file options for BuildingBlock


andrewtarzia:
The minimum working example at the bottom shows clearly that if you `write` and then `init_from_file`, you will lose bond order information. This is because PDB files do not contain bond order information in their "CONECT" records.

Therefore, I suggest removing it as an init function (the `with_structure_from_file` method is completely fine because it just loads the coordinates). Perhaps a warning is added for 6 months before removing?

What do you think, @lukasturcani?

```python

import stk
from collections import Counter

pdb_writer = stk.PdbWriter()

bb = stk.BuildingBlock(
    'CN(c1ccc(cc1)N=Cc2cccnc2)c3nc(nc(n3)N(C)c4ccc(cc4)N=Cc5ccccn5)'
    'N(C)c6ccc(cc6)N=Cc7ccccn7'
)

# Get bond order information.
orders = []
for bond in bb.get_bonds():
    orders.append(bond.get_order())

# Write to PDB.
bb.write(f'first.pdb')

# Read from PDB.
bb2 = stk.BuildingBlock.init_from_file('first.pdb')

# Get bond order information.
orders2 = []
for bond in bb2.get_bonds():
    orders2.append(bond.get_order())

print(Counter(orders), Counter(orders2)) ## >> Counter({1.0: 72, 2.0: 24}) Counter({1.0: 96})

# Write to PDB.
bb.write(f'second.pdb')

# Compare text of PDBs.
s1 = pdb_writer.to_string(bb)
s2 = pdb_writer.to_string(bb2)

print(s1 == s2)


```

lukasturcani:
I dunno, I feel like both of these methods are behaving correctly. This feels like a case where I don't want the potential user-error to detract from convience of having these two methods around. Is there a way pdb can ecnode bond orders?

lukasturcani:
OK, you've convinced me, partially. I am happy to not `write` PDB files but I think we should continue to read them. But the docs of the PDB reader should have a warning telling users that these files do not encode bond information.

lukasturcani:
Actually, I take that back. Maybe it is for the best to remove them altogether.

andrewtarzia:
Not exactly the point - I am only concerned about the `BuildingBlock.init_from_file` method. All other methods use the `Bond` attributes in the stk molecule already, which are correct. It is the creation of a building block and its bonds from a pdb file that will be "unexpected". That said, the user could already know that.

[...]

PR #449, opened 2022-05-22
Add deprecation notice for `pdb` reading

lukasturcani:
Related Issues: #446 

Removes the note that users can load `pdb` files from the docstring of 
`init_from_file`, in order to stop new users from using these files for loading
molecules. Loading will still work for now, but users will receive a `FutureWarning`
telling them that they should switch to `.mol` files.

The reason this deprecation is necessary is that `pdb` files do not encode bond
orders and the loaded building blocks may therefore have wrong or unexpected
connectivity. By explicitly not supporting this file type we make sure our users
carefully consider the bonding of their molecules and avoid surprises.

### DIFF
--- a/src/stk/molecular/molecules/building_block.py
+++ b/src/stk/molecular/molecules/building_block.py
@@ -14,6 +14,7 @@ from functools import partial
 
 import numpy as np
 import rdkit.Chem.AllChem as rdkit
+import warnings
 import vabene
 
 from ...utilities import OneOrMany, flatten, remake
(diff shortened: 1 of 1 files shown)


andrewtarzia (approved):

andrewtarzia:
Does this include the 6 month reminder to remove?

lukasturcani:
What would such a reminder look like?

andrewtarzia:
No idea - just so we do not forget in 6 months?

lukasturcani:
In practice, the removal often happens in more than six months. But I don't consider that a big deal

andrewtarzia:
Agreed.

Issue #424, opened 2022-03-04
Bad alignment in polymer topology.


andrewtarzia:
Test code below results in two weirdly aligned terminal groups with overlapping bonds.

```python
import stk

li = stk.BuildingBlock('C1=CC(=CC(=C1)Br)N', [stk.BromoFactory()])
lk = stk.BuildingBlock('C1C(CC1Br)Br', [stk.BromoFactory()])
cr = stk.BuildingBlock('C1=CC(=CC=C1Br)Br', [stk.BromoFactory()])
polymer = stk.ConstructedMolecule(
    stk.polymer.Linear(
        building_blocks=(li, lk, cr),
        repeating_unit="ABCBA",
        num_repeating_units=1
    )
)
polymer.write('lig_unopt.mol')

```

andrewtarzia:
Started a PR on this (#455).

Comments:
- The issue is with the Terminal vertex, because the "ditopic" building blocks are well aligned.
- This is because the terminal vertices were actually being converted to unaligning vertices
- Which is because, on line 667 in `linear.py`, we check for the number of placers being 1 and if so, we make it an unaligning vertex.
- However, there is no reason alignment in TerminalVertex needs more than one placer
- So the solution is to not use the unaligning vertex if the vertex is a terminal (unaligning is for non-terminal vertices).
- The code change starts on line 679 of `linear.py` and basically rejects the use of the unaligning vertex if the vertex id is in the head or tail vertex ids.

andrewtarzia:
The issue becomes more complicated because a TerminalVertex is treated like a Linear vertex if it has more than 1 functional group, leading to a bad division by zero for a 2FG, but single placer atom Building block.

PR #455, opened 2022-06-10
Fix bad alignment of terminal vertices with one placer id.

andrewtarzia:
Related Issues: #424 

The user will experience no API change. The result is better alignment of terminal building blocks in linear polymers when the building blocks have one placer id. Ideally, the user should not have to set the placer ids to get good alignment.

This change avoids the use of the UnaligningVertex when a building block has one placer id but is also a terminal vertex.

The position matrices of the polymers must change in some cases - at least test `polymer_linear3` changes.

My process of fixing is described in the associated issue.

Update: currently, the tests will fail due to the structure of  `polymer_linear3`


andrewtarzia:
Update: A bug fix for handling different types of building blocks was required. Lucky the tests cover so much chemical space!

lukasturcani (approved):

PR #458, opened 2022-06-17
Add type hints

lukasturcani:

Refactored the key makers to explicitly provide type hints rather
than rely on the default implementation. It is sometimes useful to
be able to rely on the more specific type of the key.


andrewtarzia (approved):
Makes sense to me after discussion with @lukasturcani in Discord.

PR #468, opened 2022-08-26
Automatically rebuild test image

lukasturcani:
Adds a cron job which will automatically rebuild the docker image. The `environment.yml` is also deleted because I want to rebuilt docker image to always use the latest packages, to ensure that updates to dependencies do not cause breakage with `stk`.


lukasturcani:
Closing because I can't delete the environment file because `readthedocs` needs it and I'm not sure how to commit it to the repo easily in an automated way. This can be revisited when I feel like thinking about this.

PR #478, opened 2022-12-09
Implement initialization with pathlib.Path

austin-mroz:
``xyz``, ``pdb``, and ``mdl_mol`` can now be initialized using pathlib.Paths.

Updated the type hints to allow pathlibs and imported pathlib


andrewtarzia (approved):
Big fan, no more `str(path / new_path)`

lukasturcani (approved):
Amazing! Thank you for this!

Issue #431, opened 2022-04-12
Idea for a visualisation of topologies in docs and for topology generation


andrewtarzia:
The software would take a ToplogyGraph class from stk and draw:

- spheres at each vertex, labelled or coloured by Vertex class and ID (other information can be added to label)
- lines at each edge, again labelled by ID and probably periodicity
- Each FG or connection should be labelled to
- an XYZ axis somewhere

A good feature is if there was a way to produce very nice images of these topologies for abstract figures in papers/docs. 

I imagine an interface that takes a `TopologyGraph` object.

This could be accessed as a link in the stk docs for each topology graph - similar to the current chemical structure examples. 

The main end goal is a GUI that lets you write a TopologyGraph class and draws it - in real time or via an "update" button. Therefore, making the topology definition for a user much much easier.

The main concern is for topologies like linear polymers, where the graph is defined by the class init. The user of this visualiser must have access to this as well to be able to draw a linear polymer of whatever type they are keen on.

PR #495, opened 2023-06-26
Update generic structures for topology graphs using single atom representations.

andrewtarzia:
Related Issues: #431 

This PR aims to improve the representation of topology graphs by removing artificial chemistry and focussing on the assignment and connectivity of vertices. Hopefully this improves website load times and accessibility to users. All changes are made in the doc strings, and small changes to readability are also attempted for all topology graphs.

- [ ] Add UNICODE colour maps between vertices and text.
- [ ] Finish cages
- [ ] Do for COFs
- [ ] Do for examples of macrocycles, host_guests, complexes and polymers.

### DIFF
--- a/src/stk/_internal/topology_graphs/cage/four_plus_six.py
+++ b/src/stk/_internal/topology_graphs/cage/four_plus_six.py
@@ -16,23 +16,36 @@ class FourPlusSix(Cage):
     """
     Represents a cage topology graph.
 
-    Unoptimized construction
+    Vertex connectivities:
 
     .. moldoc::
 
         import moldoc.molecule as molecule
         import stk
 
-        bb1 = stk.BuildingBlock(
-            smiles='NC1CCCCC1N',
-            functional_groups=[stk.PrimaryAminoFactory()],
+        three_c_bb = stk.BuildingBlock(
+            smiles="[Br][C]([Br])[Br]",
+            position_matrix=[
+                [-2, 0, -1],
+                [0, 0, 1],
+                [0, -2, -1],
+                [2, 0, 1],
+            ],
+            functional_groups=(stk.BromoFactory(placers=(0, 1)),),
         )
-        bb2 = stk.BuildingBlock(
-            smiles='O=Cc1cc(C=O)cc(C=O)c1',
-            functional_groups=[stk.AldehydeFactory()],
+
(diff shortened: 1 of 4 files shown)


lukasturcani:
1. do you think it is preferable to have the linkers straight?
image

lukasturcani:
Other than that -- barring the change from `COLOR1` and `COLOR2` into some kind of color unicode square (if possible) -- LGTM

lukasturcani (changes requested):
LGTM bar comments above

andrewtarzia:
> 1. do you think it is preferable to have the linkers straight?
>    image

So, the vertices are a single bead, which means that if there is a 'bend' it is because of the geometry of the input topology graph.

This is actually most clear in the M24L48 topology graph, where I clearly did not get the math right on the vertex positions. A negligible effect when the building blocks are full chemical structures though.

image

lukasturcani:
Interesting -- i guess leave it then? I assume we dont wanna mess with the topology graph vertices

lukasturcani:
Also -- for me -- i do like the fact  that there was a non-abstract example of a real cage. It kind of felt more "real" to me. Do you think having both a real and an abstract example is worthwhile? Is it worth checking with other peeps which representation they would find most useful?

andrewtarzia:
> Interesting -- i guess leave it then? I assume we don't wanna mess with the topology graph vertices
I actually intend to fix this in the future, because ultimately, these small changes will not effect people's work because geometry optimisations should be done after. But future versions should have more symmetrical inputs, it just makes more sense to me.



> Also -- for me -- i do like the fact that there was a non-abstract example of a real cage. It kind of felt more "real" to me. Do you think having both a real and an abstract example is worthwhile? Is it worth checking with other peeps which representation they would find most useful?

I think both have their place, but having both may be too much.  Perhaps we ask the discord, and maybe you can ask people in the Jelfs group?

andrewtarzia:
Also, I think the ball and stick mode reminds users that they are not limited to any particular chemistry.

lukasturcani:
Yeah -- go ahead and ask on discord and I will press people in person when I see them

andrewtarzia:
Based on discussions with others, and difficulty in making useful maps between vertex number and atom colours, I have decided to close this PR and issue.

Issue #515, opened 2023-10-20
Allow `pathlib.Path` in initialisers


andrewtarzia:
Hindsight, we should have made this happen here: #478

But `stk.BuildingBlock.init_from_file` should take a `pathlib.Path` or `str`.

Issue #494, opened 2023-06-26
Fix links


lukasturcani:
https://stk.readthedocs.io/en/latest/_autosummary/stk.Alcohol.html
![image](https://github.com/lukasturcani/stk/assets/17861210/2ce506a2-eb94-46a3-b697-596b729fd800)

andrewtarzia:
This is completed in the latest stable docs. Check and close @lukasturcani ?

lukasturcani:
interesting --- i wonder if its due to an update in sphinx? i dont think i / anyone else fixed this manually?

andrewtarzia:
No idea, but it seems solved to me?

lukasturcani:
yeah same -- will close

mostly worried that its not fixed everywhere but if so we can just open a new issue

Issue #477, opened 2022-12-01
Issue with using tuples in MetalComplex init for metals or ligands


andrewtarzia:
This one I do not understand, but the below example (from the docs) shows the issue:

```python

import stk

bb1 = stk.BuildingBlock(
    smiles="[Fe+2]",
    functional_groups=(
        stk.SingleAtom(stk.Fe(0, charge=2)) for i in range(6)
    ),
    position_matrix=[[0, 0, 0]],
)
bb2 = stk.BuildingBlock(
    smiles=(
        "C1=CC2=CC3=CC=C([N]3)C=C4C=CC"
        "(=N4)C=C5C=CC(=N5)C=C1[N]2"
    ),
    functional_groups=[
        stk.SmartsFunctionalGroupFactory(
            smarts="[#6]~[#7]~[#6]",
            bonders=(1,),
            deleters=(),
        ),
    ],
)


# Passes.
complex = stk.ConstructedMolecule(
    topology_graph=stk.metal_complex.Porphyrin(
        metals=bb1,
        ligands=bb2,
    ),
)

# Fails with: IndexError: tuple index out of range
complex = stk.ConstructedMolecule(
    topology_graph=stk.metal_complex.Porphyrin(
        metals=(bb1,),
        ligands=(bb2,),
    ),
)

```

I think it is to do with what happens after line 762 in `metal_complex.py`

Issue #187, opened 2020-07-02
Add scale parameter to topology graphs


lukasturcani:
It would be nice if the user could fine-tune the scale of a topology graph on a per instance basis. For example

```python

import stk

bb = stk.BuildingBlock('BrCCBr', [stk.BromoFactory()])
polymer = stk.ConstructedMolecule(
    topology_graph=stk.polymer.Linear(
        building_blocks=(bb, ),
        repeating_unit='A',
        num_repeating_units=2,
        scale=0.5,
    ),
)

```

PR #517, opened 2023-10-24
This PR will be a mix and match of small changes to docs and code with some fixes.

andrewtarzia:
Related Issues: #155 #187 #477 #481 #515 #494 #514

As in the title, many small changes and fixes.

- [x] Development installation process.
- [x] fix pathlib issues and links
- [x] New scale parameters
- [x] Typing in `Cof.py` and inheriting classes, and `nrotaxane.py`. This requires some help and explanation.
- [x] Handle the inheritance of `self._scale_multiplier` in Cage topologies.
- [x] Fix type in `stk.polymer.Linear.html`: `orientations [list][[float] – For each character in the repeating unit, a value between 0 and 1 (both inclusive) must be given in a [tuple]` (**tuple or list?**)

Questions:

- [x] In the new `Topology Graph Components` docs page - do we want to add all the different classes of vertex and edge?
- [x] Happy with the solution #477 to be removal of the `tuple` option because it makes far less sense for metal complexes than for cages to expect `stk` to figure out where to place building blocks.


src/stk/_internal/topology_graphs/cof/cof.py
```diff
@@ -605,7 +630,7 @@ def _get_lattice(self, vertex_alignments):
 
         xdim, ydim, zdim = (range(dim) for dim in self._lattice_size)
         # vertex_clones is indexed as vertex_clones[x][y][z]
-        lattice = [[[{} for _ in zdim] for _ in ydim] for _ in xdim]
+        lattice: list = [[[{} for _ in zdim] for _ in ydim] for _ in xdim]
```
lukasturcani, line 633: is this type hint necessary?

lukasturcani (approved):

Issue #541, opened 2025-07-14
STK overrides matplotlib backend


Luthaf:
Running the following code:

```python
import matplotlib

print(matplotlib.get_backend())

import stk

print(matplotlib.get_backend())
```

prints 

```
macosx
agg
```

on the command line, and 

```
'module://matplotlib_inline.backend_inline'
agg
```

in jupyter. This means that after importing stk in a notebook we can no longer visualize any matplotlib plots inline.

I could not find any code in this repo that would do this change, but the change only happen after importing stk itself, and not any of its dependencies (tested with the script below):

<details>

```py
import importlib
import matplotlib

DEPS = [
    "pytz",
    "vabene",
    "tzdata",
    "six",
    "pyparsing",
    "ppft",
    "pox",
    "polars",
    "PIL",  # "pillow",
    "packaging",
    "numpy",
    "networkx",
    "kiwisolver",
    "fontTools",
    "dns",  # "dnspython",
    "dill",
    "cycler",
    "scipy",
    "rdkit",
    "dateutil",  # "python-dateutil",
    "pymongo",
    "multiprocess",
    "contourpy",
    "pathos",
    "pandas",
    "matplotlib",
    "atomlite",
    "seaborn",
    "mchammer",
    "spindry",
    "stk",
]


print(matplotlib.get_backend())

for dep in DEPS:
    importlib.import_module(dep)
    print("after ", dep, " ", matplotlib.get_backend())

```

</details>

ceriottm:
Found it 

https://github.com/lukasturcani/stk/blob/fd2068e3477560d34d286e6db6f7668f84219919/src/stk/_internal/ea/plotters/selection.py#L17

https://github.com/lukasturcani/stk/blob/fd2068e3477560d34d286e6db6f7668f84219919/src/stk/_internal/ea/plotters/progress.py#L10

andrewtarzia:
Hey @Luthaf and @ceriottm ,

There should be a new release on pypi - please test that it fixes the issue, and then let me know.

Thank you for noticing this,
Andrew

ceriottm:
I confirm this fixes the issue for me. Thanks for the fast reaction.

PR #542, opened 2025-07-16
Fix matplotlib backend

andrewtarzia:
Related Issues: #541 

As mentioned in the issue, this PR just removes the setting of the matplotlib backend, which we presume is due to an out-dated decision for the evolutionary algorithm.

The output of:

```python
>>> import matplotlib
>>> print(matplotlib.get_backend())

>>> import stk
>>> print(matplotlib.get_backend())
```
 
is now (on pop-os):
```
tkagg
tkagg
```

This PR comes with a bunch of small changes associated with `ruff`, and an rdkit version constraint to avoid issues with metal-containing smiles strings.


src/stk/_internal/utilities/updaters/pdb.py
```diff
@@ -1,38 +1,39 @@
...
-    Returns
-    -------
-    :class:`.Molecule`
+    Returns:
         The molecule.
 
     """
 
-    molecule = remake(
+    # Handle API change, failure gives None.
```
lukasturcani, line 23: remove comment

lukasturcani (approved):

PR #543, opened 2025-10-02
Use uv in stk tooling

andrewtarzia:

I am going through and trying to fix a few small issues with stk and I thought I would test this out.

Happy to ignore if you think uv is not the right move here.

I expect to submit more PRs in quick succession, if you are ok with letting mypy fail here. In particular, I want to unpin the rdkit version.


lukasturcani (approved):
LGTM, ty!

Issue #538, opened 2025-03-01
How to handle changes in rdkit ETKDG algorithm in latest versions, that slightly change stk output?


andrewtarzia:
Maybe a discussion more than anything, but basically, new versions of rdkit change the etkdg algorithm (we should check if stable), which cause stk tests to fail - mostly position based.

A detailed analysis of safe versions are necessary before going forward.

lukasturcani:
if its just a postiion change i would just update tests to use the version of rdkit

Issue #487, opened 2023-05-31
Add support for atomlite


lukasturcani:
by making converstions to entries

andrewtarzia:
I have done this here: https://cgexplore.readthedocs.io/en/latest/_autosummary/cgexplore.utilities.AtomliteDatabase.html

Issue #54, opened 2019-08-13
units should be included with all values in doc strings


andrewtarzia:
For example, in the CageWithGuest API, there are no units given for the angle. 

This is a minor problem but would make the code much more user-friendly.

lukasturcani:
Can you add a specific list please, so that I know what to tick off?

lukasturcani:
I'm gonna reopen this if you add the list.

PR #544, opened 2025-10-03
Remove rdkit pin, fix tests and errors.

andrewtarzia:
Related Issues: #538 

Removing this pin will alter the smiles strings of any dative-bond (`->` or `<-`) containing species (from: https://github.com/rdkit/rdkit/releases?q=2024_09_&expanded=true)

> When writing SMILES, organic subset atoms which are bonded to "metals" will
> always be written in square brackets, i.e. with their H count explicit. Here
> the definition of "metal" is any atom matching an "M" query (the corresponding
> SMARTS is [!#0!#1!#2!#5!#6!#7!#8!#9!#10!#14!#15!#16!#17!#18!#33!#34!#35!#36!#52!#53!#54!#85!#86])

Importantly, it does not alter the construction outcome (or mol file writing), just the writing of smiles. But, new versions of rdkit have changed conformations slightly, so geometries in tests were needed to change too.

Things to do:

- [ ] Fix the tests with broken smiles
- [ ] Add warning that smiles output will change
- [ ] Fix the tests with changed coordinates
- [ ] Add atomlite output #487 while I am here


lukasturcani (approved):
i wiould remove the atomlite conversation functions but otherwise lgtm

